PR #7, opened 2019-01-10
Revert "android-test-lib adb_join_wifi: uninstall the package"

liuyq:
uninstall that package would cause wifi disconnected

as reported here:
https://projects.linaro.org/browse/LSS-402

This reverts commit 1136333f3659ab251096848444d94ae7e478d31a.


chase-qi:
Sanity check passed for the PR(as no file get tested) but failed for the push event, refer to https://github.com/Linaro/test-definitions/runs/48484941

https://github.com/Linaro/test-definitions/pull/8 will fix the issues.

PR #8, opened 2019-01-11
Improve sanity check

chase-qi:
* Use 'git diff --name-only HEAD~1' to list changed files
* Add instructions on the installation of a specific version shellcheck
* Add descriptions for the skipped pycodestyle and shellcheck codes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,9 @@ python:
   - "3.6"
 
 # Pre-installed packages.
-# shellcheck 0.5.0
+# shellcheck v0.5.0
+# Refer to the following link to install specific version shellcheck binary.
+# https://github.com/koalaman/shellcheck#installing-the-shellcheck-binary
 # php 7.2
 install:
   - pip install pyyaml pycodestyle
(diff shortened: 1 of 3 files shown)


danrue:
'git diff --name-only HEAD~1' - so are we going to require only one commit per pull request?

How hard would it be to fix all of the issues so that we can run it against the whole repository every time?

mwasilew:
We don't want it on the whole repo because there are items that fail and we know about that. The option @chase-qi changed only applies to -g. You can still run validate.py on the whole repo but it will show a lot of failed checks.

chase-qi:
@danrue Fixing all of the existing issues isn't too hard but takes time, I will take a look at some point. At least for the moment, we want to test the changed files only so that we can focus on the PR itself. There are some means to get changed files for multiple commits in one PR. I will test them.

Hard to understand, but probably will work for both Travis and local run? 

```bash
git diff --name-only FETCH_HEAD $(git merge-base FETCH_HEAD master)
```

Travis only:
```bash
git diff --name-only $TRAVIS_COMMIT_RANGE
```

PR #13, opened 2019-02-04
android/tradefed: improve reboot variable comparison

mwasilew:
Make sure that comparison is case insensitive to catch any possible
reboot in LAVA. Also align the comparison condition between different
tradefed implementations.


dave-pigott:
Looks good. Tested for both template types and it passes.

PR #15, opened 2019-02-11
Add MultiNode TradeFed tests for --shards and with session retry

kateyy:
Continuation of changes previously discussed on Gerrit: [tradefed_shards_with_retry](https://review.linaro.org/#/q/topic:tradefed_shards_with_retry+(status:open+OR+status:merged))

### DIFF
--- a/automated/android/multinode/tradefed/example-job-template-cts.yaml
+++ b/automated/android/multinode/tradefed/example-job-template-cts.yaml
@@ -1,0 +1,204 @@
+job_name: MutliNode_xTS_template
+timeouts:
+  job:
+    hours: 2
+priority: medium
+visibility: public
+reboot_to_fastboot: false
+
+protocols:
+  lava-lxc:
+    master:
+      name: lxc-xts-master
+      template: debian
+      distribution: debian
+      release: stretch
+    worker:
+      name: lxc-xts-worker
+      template: debian
+      distribution: debian
+      release: stretch
+  lava-multinode:
+    # There must be one master and an arbitrary number of additional workers, so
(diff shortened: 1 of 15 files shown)


kateyy:
What is the policy to fix W503 and W504; do you prefer to violate a maximum line length? Why not using black formatting as for LAVA itself? (Which is not fully compatible with pycodestyle).

automated/android/multinode/tradefed/example-job-template-cts.yaml
```diff
@@ -0,0 +1,204 @@
...
+
+- test:
+    namespace: tlxc
+    role:
+    - master
+    - worker
+    timeout:
+      minutes: 20
+    definitions:
+    - repository: https://review.linaro.org/qa/test-definitions
```
chase-qi: Can you please replace repository url with https://github.com/Linaro/test-definitions ?

chase-qi:
Sanity check is line length flexible already https://github.com/Linaro/test-definitions/blob/master/sanity-check.sh#L7 . Introducing black formatting needs more investigation, I don't know if LAVA using it. @mwasilew ideas/comments? I personally hope we are more transparent, less solid black.

And please ignore the failure with job example. We need some magic to distinguish between test job and test definition yaml files. Tracking with https://github.com/Linaro/test-definitions/issues/17

mwasilew:
@chase-qi lava uses 'black' formatting now. In our case I see not much benefit to be honest but I have nothing against moving to this code style.

chase-qi:
Again, thanks for contributing. I am merging the PR.

PR #16, opened 2019-02-12
MultiNode Tradefed: Refactoring and userdata reset

kateyy:
This branch is on top of #15 - please review the two first commits in the other pull request first.

New (optional) feature added here: After a Tradefed session finished and before triggering a `--retry` run, boot all DUTs into fastboot and flash a userdata image to reset them to a clean state. This reduces side effects some (CTS) test module have on others, when they don't undo changes that happen during test executions. Finally, this helps to produce more reliable test results with a lower number of retry runs.

This PR also contains a list of cleanup/refactoring commits. Let me know if you prefer having them in a separate PR.


chase-qi:
Hi @kateyy, commit 'Add MultiNode TradeFed tests for --shards and with session retry' has been merged already, could you rebase the PR against master? And please fix the shellcheck https://travis-ci.com/Linaro/test-definitions/builds/101075823#L221

kateyy:
> Hi @kateyy, commit 'Add MultiNode TradeFed tests for --shards and with session retry' has been merged already, could you rebase the PR against master? And please fix the shellcheck https://travis-ci.com/Linaro/test-definitions/builds/101075823#L221

Done.

automated/lib/android-test-lib
```diff
@@ -33,8 +33,6 @@ adb_debug_info() {
 }
 
 initialize_adb() {
-    # enable debug
-    set -x
```
chase-qi: +1. I checked existing sigle node tradefed test definitions. We run setup.sh(. ./setup.sh) with the same context from run.sh, and LAVA runs run.sh with sh '-ex' enabled by default, means the trace option should be enabled already.

automated/android/multinode/tradefed/tradefed-multinode.yaml
```diff
@@ -83,7 +78,7 @@ run:
         # available in the CTS execution scripts.
         - |
             exec_result=0
-            sudo -u testuser env "PATH=${PATH}" ./tradefed-multinode.sh \
```
chase-qi: The testuser was introduced for vts, vts didn't work with root for some reasons. I am not sure if it still is the case now. As using root for both will make things simpler, I tend to add +1 for this commit. We can revisit it when we  introduce vts to multinode testdef.

automated/lib/py_util_lib.py
```diff
@@ -0,0 +1,45 @@
+"""Shared Python 3 utility code.
```
chase-qi: We already have `automated/lib/py_test_lib.py` for the same purpose. Is there a reason to add a new one?

chase-qi:
@kateyy ping

automated/lib/py_util_lib.py
chase-qi: Sounds good to me.

PR #19, opened 2019-02-14
noninteractive-tradefed: add more debug

mwasilew:

### DIFF
--- a/automated/android/noninteractive-tradefed/monitor_adb.sh
+++ b/automated/android/noninteractive-tradefed/monitor_adb.sh
@@ -2,6 +2,7 @@
 while true;
 do
     date
-    fastboot boot /lava-lxc/boot*.img
+    lsusb
+    adb devices
     sleep 30
 done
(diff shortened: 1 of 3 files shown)


automated/android/noninteractive-tradefed/monitor_fastboot.sh
```diff
@@ -1,4 +1,6 @@
 #!/bin/sh -x
 while true; do
```
chase-qi: Not this patch related, but looks like the script simply run 'fastboot boot'  every 30 seconds. I think we should at least check if a fastboot device appeared.

automated/android/noninteractive-tradefed/monitor_adb.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh -x
+while true; do
+date
```
chase-qi: Indentation?

automated/android/noninteractive-tradefed/monitor_fastboot.sh
mwasilew: fastboot will wait indefinitely for the device to appear. The whole idea is to boot the board if it goes to fastboot mode.

automated/android/noninteractive-tradefed/monitor_adb.sh
mwasilew: will fix, thanks.

PR #29, opened 2019-03-08
Add multinode job example

chase-qi:


chase-qi:
Once https://github.com/Linaro/test-definitions/pull/28 get reviewed/merged, sanity check will validate:
* yaml syntax for generic yaml files like job example.
* yaml syntax and required keys for test definition yaml files.

The failed chesk reported here will be fixed/skipped.

PR #32, opened 2019-03-12
Chroot

roxell:
run ltp and kselftest in a chroot


automated/linux/ltp/ltp.yaml
```diff
@@ -49,10 +49,3 @@ run:
         - cd ./automated/linux/ltp/
         - ./ltp.sh -T "${TST_CMDFILES}" -s "${SKIP_INSTALL}" -v "${LTP_VERSION}" -M "${TIMEOUT_MULTIPLIER}" -R "${ROOT_PASSWD}" -b "${BOARD}" -g "${BRANCH}" -e "${ENVIRONMENT}" -S "${SKIPFILE}"
         - ../../utils/send-to-lava.sh ./output/result.txt
-
-parse:
```
chase-qi: I think we are safe to remove the parse pattern here. Objections?

danrue:
A few initial thoughts and questions on this:
- Should the scripts be run with set -e instead of so many "|| exit"? And what about all the places where the exit code isn't being checked?
- Can the existing kselftest and ltp test definitions be re-used, rather than re-implemented here?
- "ARM32_ROOTFS" is named based on your use-case, but in fact this could be used for rootfs that are not ARM32.

In general it looks quite fragile to me and I'm wondering how it could be made simpler and easier to maintain over time.

chase-qi:
Hi Dan,

Thanks for the comments. 

> A few initial thoughts and questions on this:
> 
> * Should the scripts be run with set -e instead of so many "|| exit"? And what about all the places where the exit code isn't being checked?

+1 on enabling '-e'.

> * Can the existing kselftest and ltp test definitions be re-used, rather than re-implemented here?

It is basically sending most of the lines(params and run-steps) from the original yaml to new_root/run.sh. Overwriting params is the tricky part. After chroot, the new test shell with new_root doesn't share the original lava-test-shell env, we have to render/evaluate the vars first and then send them to new_root/run.sh so that the defaults get overwritten.

> * "ARM32_ROOTFS" is named based on your use-case, but in fact this could be used for rootfs that are not ARM32.
> 
> In general it looks quite fragile to me and I'm wondering how it could be made simpler and easier to maintain over time.

It is fragile indeed. This is dirty hack in test definitions, not ideal for long run. IMHO, if running tests in chroot env is a valid use case, it should be supported by LAVA. From my point of view, LAVA needs to support to switch to second shell prompt and apply its overlay tarball to the second rootfs.

automated/linux/ltp/ltp.yaml
nareshkamboju: @chase-qi I agree. We are not using this parse.

roxell:
I'm unsure if I think the "set -e" + "(" ")" is better than "|| exit" anyway I changed it.
I also changed the ARM32_ROOTFS -> NEW_ROOTFS

danrue:
OK! Thanks @roxell :+1:

PR #33, opened 2019-03-13
sanity-check: print installation tips when required module not installed

chase-qi:


chase-qi:
Output example:

```
$ ./validate.py -h
No module named 'magic'
Install python-magic: pip3 install python-magic
```

PR #34, opened 2019-03-13
linux/perf: Adding perf test case for LKFT test plan

nareshkamboju:
Perf is a profiler tool for Linux and is used to collect and analyze
performance and trace data.
This test runs following perf commands:
- record(record events for later reporting)
- report(break down events by process, function, etc)
- stat(obtain event counts)
- test(Run sanity tests)

Upstream link,
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/tools/perf/Documentation/perf.txt

Ref:
Test perf in LKFT
https://projects.linaro.org/browse/KV-195

### DIFF
--- a/automated/linux/perf/perf.sh
+++ b/automated/linux/perf/perf.sh
@@ -1,0 +1,112 @@
+#!/bin/sh
+
+# shellcheck disable=SC1091
+. ../../lib/sh-test-lib
+OUTPUT="$(pwd)/output"
+RESULT_FILE="${OUTPUT}/result.txt"
+export RESULT_FILE
+SKIP_INSTALL="false"
+# List of test cases
+TEST="record report stat test"
+# PERF version
+PERF_VERSION="$(uname -r | cut -d . -f 1-2)"
+
+usage() {
+    echo "Usage: $0 [-s <true|false>]" 1>&2
+    exit 1
+}
+
+while getopts "s:h" arg; do
+   case "$arg" in
+     s) SKIP_INSTALL="${OPTARG}";;
+     h|*) usage ;;
+   esac
+done
+
+# Run perf record tests
+run_perf_record() {
+    # Test 'perf record'
+    info_msg "Performing perf record test..."
+    TCID="perf_record_test"
+    perf record -e cycles -o perf-lava-test.data ls -a  2>&1 | tee perf-record.log
+    samples=$(grep -ao "[0-9]\\+[ ]\\+samples" perf-record.log| cut -f 1 -d' ')
+    if [ "${samples}" -gt 1 ]; then
+        report_pass "${TCID}"
+    else
+        report_fail "${TCID}"
+    fi
+    rm perf-record.log
+}
+
+# Run perf report tests
+run_perf_report() {
+    # Test 'perf report'
+    info_msg "Performing perf report test..."
+    TCID="perf_report_test"
+    perf report -i perf-lava-test.data 2>&1 | tee perf-report.log
+    pcnt_samples=$(grep -c -e "^[ ]\\+[0-9]\\+.[0-9]\\+%" perf-report.log)
+    if [ "${pcnt_samples}" -gt 1 ]; then
+        report_pass "${TCID}"
+    else
+        report_fail "${TCID}"
+    fi
+    rm perf-report.log perf-lava-test.data
+}
+
+# Run perf stat tests
+run_perf_stat() {
+    # Test 'perf stat'
+    info_msg "Performing perf stat test..."
+    TCID="perf_stat_test"
+    perf stat -e cycles ls -a 2>&1 | tee perf-stat.log
+    cycles=$(grep -o "[0-9,]\\+[ ]\\+cycles" perf-stat.log | sed 's/,//g' | cut -f 1 -d' ')
+    if [ -z "${cycles}" ]; then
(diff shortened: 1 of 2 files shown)


nareshkamboju:
Perf test is been tested on LAVA LKFT instance on x86_64,
https://lkft.validation.linaro.org/scheduler/job/639333#L4485

automated/linux/perf/perf.sh
```diff
@@ -0,0 +1,112 @@
...
+else
+    dist_name
+    # shellcheck disable=SC2154
+    case "${dist}" in
+      debian|ubuntu)
+	pkgs="linux-tools-${PERF_VERSION}"
+        install_deps "${pkgs}" "${SKIP_INSTALL}"
+        ;;
+      centos|fedora)
+	pkgs="perf"
```
mwasilew: how do you control version of perf on centos?

nareshkamboju: I do remember the fact that. perf version matching with kernel version is always a challenge.
Currently Open embedded recipe is getting pre-built the one i have tested is 4.20 is working.

nareshkamboju: @mwasilew I have not tested perf install on centos.
I have checked on redhat page and found yum install perf is the way to install
# yum install perf 
ref link,
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux_for_real_time/7/html/tuning_guide/about_perf

mwasilew: @nareshkamboju my point here is that for debian based systems you try to install perf package matching running kernel. For RPM systems you install version that is 'latest'. IIRC yum can forcibly install version you require: yum install <package name>-<version info>

chase-qi: I just checked in centos docker container. We can install particular package version in the format Milosz mentioned above.  You will need to check on centos for detecting kernel or package version. Within the container, perf package version doesn't match my host kernel, that is expected.

```
[root@ccf8f43bc871 /]# yum list perf --showduplicates
Loaded plugins: fastestmirror, ovl
Loading mirror speeds from cached hostfile
 * base: mirrors.aliyun.com
 * extras: mirrors.aliyun.com
 * updates: mirrors.aliyun.com
Available Packages
perf.x86_64                                                                                                  3.10.0-957.el7                                                                                                       base   
perf.x86_64                                                                                                  3.10.0-957.1.3.el7                                                                                                   updates
perf.x86_64                                                                                                  3.10.0-957.5.1.el7                                                                                                   updates
[root@ccf8f43bc871 /]# yum install perf-3.10.0-957.el7
```

automated/linux/perf/perf.yaml
```diff
@@ -0,0 +1,29 @@
+metadata:
+    name: perf
+    format: "Lava-Test-Shell Test Definition 1.0"
+    description: "Perf is a profiler tool for Linux and is used to collect and analyze performance and trace data.
+                  This test runs following perf commands: record(record events for later reporting), report(break
+                  down events by process, function, etc), stat(obtain event counts) and test(Run sanity tests)."
+    maintainer:
+        - naresh.kamboju@linaro.org
+        - milosz.wasilewski@linaro.org
+    os:
```
chase-qi: Add debian and centos?

automated/linux/perf/perf.sh
```diff
@@ -0,0 +1,112 @@
...
+        install_deps "${pkgs}" "${SKIP_INSTALL}"
+        ;;
+      *)
+        warn_msg "Unsupported distribution: package install skipped"
+    esac
+fi
+
+# List of test cases "record report stat test"
+for tests in ${TEST}; do
+    run_perf_"${tests}"
```
chase-qi: Because the test script skip package installation by default, I guess mainly for OE based builds, so we need to either enable 'sh -e' or 'which perf || exit'.

From my point of view, we should set SKIP_INSTALL to false by default, but it is your call.

nareshkamboju: SKIP_INSTALL changed to false.
However my test job have parameter to set it to 'true'.

"which perf || exit" to check pre-installed or not.
I will update that part.

automated/linux/perf/perf.yaml
nareshkamboju: Added debian.
I am thinking if we do not find a way to install perf package on centos. Shall we not include centos here.

automated/linux/perf/perf.sh
```diff
@@ -0,0 +1,112 @@
...
+info_msg "Output directory: ${OUTPUT}"
+
+if [ "${SKIP_INSTALL}" = "True" ] || [ "${SKIP_INSTALL}" = "true" ]; then
+    info_msg "install perf skipped"
+else
+    dist_name
+    # shellcheck disable=SC2154
+    case "${dist}" in
+      debian|ubuntu)
+	pkgs="linux-tools-${PERF_VERSION}"
```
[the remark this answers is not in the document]
nareshkamboju: This is really useful information.
I have changed the package name to linux-perf-<kernel-version>

automated/linux/perf/perf.yaml
chase-qi: If it is not needed/tested on centos, then lets don't claim it supports centos. Then we should drop the installation lines for centos as well.

PR #37, opened 2019-03-20
Use lava-test-set in automated/linux/piglit

alimon:


alimon:
@chase-qi I added other commit to fix pycodestyle checks.

PR #53, opened 2019-04-11
linux: add test definition for bootrr

chase-qi:

### DIFF
--- a/automated/linux/bootrr/bootrr.yaml
+++ b/automated/linux/bootrr/bootrr.yaml
@@ -1,0 +1,32 @@
+metadata:
+    format: Lava-Test Test Definition 1.0
+    name: bootrr
+    description: "Run bootrr https://github.com/andersson/bootrr"
+    maintainer:
+        - milosz.wasilewski@linaro.org
+        - chase.qi@linaro.org
+    os:
+        - debian
+        - ubuntu
+        - centos
+        - fedora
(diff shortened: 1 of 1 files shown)


chase-qi:
@ndechesne does this look right for you?

chase-qi:
> @ndechesne does this look right for you?

automated/linux/bootrr/bootrr.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+params:
+    # Refer to https://github.com/andersson/bootrr/tree/master/boards for boards supported.
+    BOARD: "arrow,apq8096-db820c"
+    SKIP_INSTALL: false
+
+run:
+    steps:
+        - . ./automated/lib/sh-test-lib
+        - install_deps git "${SKIP_INSTALL}"
```
chase-qi: With the current approach, install_deps step can be skipped, but the following 'git clone' is required. Checking if git installed is redundant as lava run the 'run steps' with sh '-e' enabled by default. We also can do:

Add a param for existing bootrr(built-in or overlay applied somehow), when the param defined, skip both install_deps and git clone, then use the existing bootrr directory directly.

chase-qi: I am merging this PR. @alimon will send another PR to add a param for using existing bootrr.

PR #59, opened 2019-04-24
kvm-unit-tests: Adding DEVICE parameter for juno-r2

nareshkamboju:
Juno-r2 have big.LITTLE processors  2 x A57 and 4 x A53.
When KVM runs migrated from little to big or vice versa the KVM unit test
got failed. To Avoid the task migration we have to offlines big processors.

This is a complete work around to get kvm unit test runs on juno-r2.

### DIFF
--- a/automated/linux/kvm-unit-tests/kvm-unit-tests.sh
+++ b/automated/linux/kvm-unit-tests/kvm-unit-tests.sh
@@ -1,21 +1,25 @@
 #!/bin/sh
-set -x
+
 # shellcheck disable=SC1091
 . ../../lib/sh-test-lib
+# shellcheck disable=SC1091
+. ./setup.sh
 
 OUTPUT="$(pwd)/output"
 RESULT_FILE="${OUTPUT}/result.txt"
 RESULT_LOG="${OUTPUT}/result_log.txt"
 SKIP_INSTALL="false"
+DEVICE=""
 
 usage() {
-    echo "Usage: $0 [-s <true|false>]" 1>&2
+    echo "Usage: $0 [-s <true|false>] [-d <juno-r2>]" 1>&2
     exit 1
 }
 
-while getopts "s:h" o; do
+while getopts "s:d:h" o; do
   case "$o" in
     s) SKIP_INSTALL="${OPTARG}" ;;
+    d) DEVICE="${OPTARG}" ;;
     h|*) usage ;;
   esac
 done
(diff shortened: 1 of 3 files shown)


automated/linux/kvm-unit-tests/kvm-unit-tests.sh
```diff
@@ -82,6 +84,13 @@ else
   install
 fi
 
+if [ "${DEVICE}" = "JUNO-R2" ] || [ "${DEVICE}" = "juno-r2" ]; then
```
danrue: I think I see why you do this, but why not just convert DEVICE to lower case so you don't have to have multiple case cases?

automated/linux/kvm-unit-tests/kvm-unit-tests.sh
```diff
@@ -82,6 +84,13 @@ else
...
+    # fixup: hard coded
```
danrue: eh, i'd just remove this comment.

danrue:
Looks good to me @nareshkamboju besides the little nit picks I found. :+1:

automated/linux/kvm-unit-tests/kvm-unit-tests.sh
nareshkamboju: @danrue +1 Done.

automated/linux/kvm-unit-tests/kvm-unit-tests.sh
nareshkamboju: @danrue only checks for lower case "juno-r2".

mwasilew:
Honestly I don't like this. There are a lot of assumptions here that are only true for Linaro LAB. Is CPU enumeration always the same? IIRC it had changed in the past. I would rather see this as a 'setup.sh' script specific to device and called based on the parameters set in YAML. Default implementation should be no-op.

nareshkamboju:
Hi Milosz,

Thanks for the review.

> Honestly I don't like this. There are a lot of assumptions here that are only true for Linaro LAB. Is CPU enumeration always the same? IIRC it had changed in the past. I would rather see this as a 'setup.sh' script specific to device and called based on the parameters set in YAML. Default implementation should be no-op.

As per your review comments, I have added setup.sh script to detect a57 cluser cpus on juno-r2 and offline them before running tests and online those cpus after successful kvm run.

I have tested this new patch on LAVA.
https://lkft.validation.linaro.org/scheduler/job/691840#L1904

nareshkamboju:
@mwasilew  This patch might not be need because,
I have done changes in test templates in config.git

Please review this config patch,
kvm-unit-tests: juno-r2: offline a57 cpus from inline
https://review.linaro.org/c/ci/job/configs/+/31241

mwasilew:
@nareshkamboju I think test plan is more appropriate place for your change.

PR #66, opened 2019-05-16
automated/android: bail in case adb root times out

mwasilew:


mwasilew:
Since there were no objections I'm merging this patch.

PR #67, opened 2019-05-17
Kselftest skips

roxell:


automated/linux/kselftest/skipfile-lkft.yaml
```diff
@@ -27,7 +27,14 @@ skiplist:
     url: https://bugs.linaro.org/show_bug.cgi?id=3297
     environments: all
     boards: all
-    branches: all
+    branches:
```
mwasilew: I don't understand this change. It seems not matching the commit message. Could you explain?

danrue: It's because he's removing -next from the list; on next, kselftest times out hung tests after 30 seconds.

PR #68, opened 2019-05-20
network-basic: skip set interface down on NFS mounted fs

nareshkamboju:
network-basic: skip set interface down on NFS mounted fs

x86 running network basic test failed to make clean exit because
test case set network interface down and NFS lost connection. so
NFS parameter introduced to skip "set interface down" command when
NFS is true.

+ echo 'ip-link-interface-down pass'
[  241.270557] nfs: server 10.66.16.116 not responding, still trying
[  241.270569] nfs: server 10.66.16.116 not responding, still trying

ref:
https://lkft.validation.linaro.org/scheduler/job/735083#L1575

### DIFF
--- a/automated/linux/network-basic/network-basic.sh
+++ b/automated/linux/network-basic/network-basic.sh
@@ -5,16 +5,18 @@
 OUTPUT="$(pwd)/output"
 RESULT_FILE="${OUTPUT}/result.txt"
 export RESULT_FILE
+NFS="false"
 
 usage() {
-    echo "Usage: $0 [-s <true|false>] [-i <interface>]" 1>&2
+    echo "Usage: $0 [-s <true|false>] [-i <interface>] [-n <true|false>]" 1>&2
     exit 1
 }
 
-while getopts "s:i:" o; do
+while getopts "s:i:n:" o; do
   case "$o" in
     s) SKIP_INSTALL="${OPTARG}" ;;
     i) INTERFACE="${OPTARG}" ;;
+    n) NFS="${OPTARG}" ;;
     *) usage ;;
   esac
 done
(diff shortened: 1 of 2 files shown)


nareshkamboju:
Adding a parameter (NFS) to skip network interface down when
NFS: true 
might be one more approach to look at this problem.

- Naresh

mwasilew:
@nareshkamboju could you set the variable in the test and use conditional statement in the test? This way you can support both NFS and non-NFS cases.

nareshkamboju:
@mwasilew  NFS parameter added.

nareshkamboju:
@mwasilew @danrue Please review this change request.

automated/linux/network-basic/network-basic.sh
```diff
@@ -50,7 +52,12 @@ run "route" "print-routing-tables"
 run "ip link set lo up" "ip-link-loopback-up"
 run "route" "route-dump-after-ip-link-loopback-up"
 run "ip link set ${INTERFACE} up" "ip-link-interface-up"
-run "ip link set ${INTERFACE} down" "ip-link-interface-down"
+# NFS mounted filesystem hangs when interface set down
+# so skip set <interface> down
+
+if [ "${NFS}" = "false" ]; then
```
mwasilew: Could you make it case insensitive:
[ "${NFS}" = "false" ] || [ "${NFS}" = "False" ]

Also please add 'skip' to the 'ip-link-interface-down' if the condition is not met

nareshkamboju: Thanks @mwasilew really valuable comment.
I have addressed your comments.

automated/linux/network-basic/network-basic.sh
```diff
@@ -50,7 +52,15 @@ run "route" "print-routing-tables"
...
 run "ip link set ${INTERFACE} up" "ip-link-interface-up"
-run "ip link set ${INTERFACE} down" "ip-link-interface-down"
+# NFS mounted filesystem hangs when interface set down
+# so skip "ip link set <interface> down" when NFS is not false
+
+if [ "${NFS}" = "false" ] || [ "${NFS}" = "False" ]; then
+    run "ip link set ${INTERFACE} down" "ip-link-interface-down"
+else
+    info_msg "Skipping ip link set ${INTERFACE} down on NFS mounted file systems"
+    echo "ip-link-interface-down skip"
```
mwasilew: We have "report_skip" from sh-test-lib for skipping tests.

nareshkamboju: oops.
I have added "report_skip" to lib.
thanks for the reminder :-)

PR #73, opened 2019-06-01
kselftest: update output parsing to work on v5.2

roxell:
kselftest changed outout format on next that will be v5.2.


roxell:
This is the output from the new output format:
https://lkft.validation.linaro.org/results/758759/2_kselftest?search=&length=100#table

mwasilew:
@roxell so this will break stable jobs that use kselftest from latest stable?

roxell:
kselftest v5.2 and later will need my PR to parse the tests. so for earlier kseftest versions we should use a released tag from test-definition in order to kseftest to parse correctly.

PR #74, opened 2019-06-10
kselftest skiplist update

nareshkamboju:
Updating kselftest skiplist for hangs test cases on linux next and mainline.
And when kselftest version update to 5.1 on stable branches
there is a hang for arm64 on 4.19 kernel branch due to test_progs

test_progs has to cherry-picked in to 2019.05
or 
create a new tag 2019.06 = 2019.05 + one patch [ 2c294af7cf ( kselftest: skip hang test case test_progs on 4.19 branch ) ]


danrue:
This looks good to me but per @roxell, I believe skiplist isn't supported on mainline/next at the moment.

PR #75, opened 2019-06-10
kselftest: update how to skip tests

roxell:
Upstream kselftest updated how to generate the run_kselftest.sh script
and that broke how we wants hack us self in to skip tests, update to
solve it for now. Binary tests that hangs for more than 30seconds will
automatically be killed by the kselftest framework. The only reminding
thing until we can drop our skiplist completely is if scripts hangs.


automated/linux/kselftest/kselftest.sh
```diff
@@ -151,7 +151,8 @@ if [ -n "${SKIPLIST}" ]; then
     # shellcheck disable=SC2086
     for test_name in ${SKIPLIST}; do
         # shellcheck disable=SC2086
-        sed -i "/.\/${test_name}/c\echo \"selftests: ${test_name} [SKIP]\"" run_kselftest.sh
+        sed -i "/${test_name}/d" run_kselftest.sh
```
danrue: Looking at run_kselftest.sh (https://git.io/fjgG2), I notice that when there are a list of tests, all but the last ones end the line with "\\". So, if the test being skipped is the last in the list, this will not work.

danrue: Also, the other case I wonder if you tested, is what if the last test was removed from a test section. Will it work?

PR #76, opened 2019-06-10
igt: print igt test result with python script

arthur-she:
Due to html2text is quite old and the code hasn't been updated since 2005, it might be concerned to be included in the image.
Since igt-gpu-tools will generate a plain text test result and save it with JSON format.
I created a python script to print it in LAVA test log.
With this approach, html2text will not be needed.

### DIFF
--- a/automated/linux/igt/print-test-result.py
+++ b/automated/linux/igt/print-test-result.py
@@ -1,0 +1,39 @@
+#!/usr/bin/python
+import argparse
+import sys
+import json
+
+
+def print_result(results):
+    try:
+        for test, content in results['tests'].iteritems():
+            print '<LAVA_SIGNAL_STARTTC %s>' % test
+            print '************************************************************************************************************************************'
+            print '%-15s %s' % ('Test:', test)
+            print '%-15s %s' % ('Result:', content['result'])
+            print '%-15s %s' % ('Command:', content['command'])
+            print '%-15s %s' % ('Environment:', content['environment'])
+            print '%-15s %s' % ('Returncode:', content['returncode'])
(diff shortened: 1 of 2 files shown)


arthur-she:
Test job
https://validation.linaro.org/scheduler/job/1917642

automated/linux/igt/print-test-result.py
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/python
+import argparse
+import sys
+import json
+
+
+def print_result(results):
+    try:
+        for test, content in results['tests'].iteritems():
+            print '************************************************************************************************************************************'
```
mwasilew: Since you have all the results and logs in one place I would try to use LAVA_SIGNAL_STARTTC and LAVA_SIGNAL_ENDTC. This way LAVA will know about the test name, result and log. I think the proper sequence should be
<LAVA_SIGNAL_STARTTC test_case_name>
log output here. You can include all "print" statements you have below.
<LAVA_SIGNAL_TESTCASE TEST_CASE_ID=test_case_name RESULT=test_result>
<LAVA_SIGNAL_ENDTC test_case_name>

I didn't try so I'm not 100% sure.

mwasilew: LAVA_SIGNAL_TESTCASE should come last. Here is the example:
https://validation.linaro.org/scheduler/job/1918367#L901

arthur-she: I've revised the script.
Here is the test job
https://validation.linaro.org/scheduler/job/1918440#L2303

mwasilew:
LGTM. LAVA seems to be slow to pick the signal but I think it might be fixed on LAVA side. If not you'll need to add some short sleep after each signal tag. Merging this now.

PR #77, opened 2019-06-11
ltp: add Schneider RZN1D board into skipfile-lkft.yaml

shawnguo2:
Add Schneider RZN1D board into skipfile-lkft.yaml to skip test cases
that are skipped for dragonboard-410c board.

Signed-off-by: Shawn Guo <shawn.guo@linaro.org>


mwasilew:
@nareshkamboju is this the correct skipfile?

danrue:
This change is OK with me, if they're looking to re-use lkft's skips for RZN1D. Let's have @nareshkamboju approve too.

PR #78, opened 2019-06-12
ltp: check unprivileged_userns_clone existence first

shawnguo2:
The unprivileged_userns_clone procfs entry exists only with distro
kernel with particular patch applied, e.g. the patch [1] with debian
kernel.  On development boards which mostly run upstream kernel without
such patch, the ltp script will give following message which looks like
something goes wrong.

++ sysctl -n kernel.unprivileged_userns_clone
sysctl: cannot stat /proc/sys/kernel/unprivileged_userns_clone: No such file or directory
+ '[' '' -eq 0 ']'
./ltp.sh: line 157: [: : integer expression expected

Let's check unprivileged_userns_clone existence first before accessing
it, so that we can avoid above "error" messages and the confusion caused
by it.

[1] https://salsa.debian.org/kernel-team/linux/blob/master/debian/patches/debian/add-sysctl-to-disallow-unprivileged-CLONE_NEWUSER-by-default.patch

Signed-off-by: Shawn Guo <shawn.guo@linaro.org>

### DIFF
--- a/automated/linux/ltp/ltp.sh
+++ b/automated/linux/ltp/ltp.sh
@@ -154,9 +154,11 @@ prep_system() {
         systemctl stop systemd-timesyncd
     fi
     # userns07 requires kernel.unprivileged_userns_clone
-    if [ "$(sysctl -n kernel.unprivileged_userns_clone)" -eq 0 ]; then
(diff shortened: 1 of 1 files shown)


mwasilew:
LGTM. @nareshkamboju could you take a look?

PR #88, opened 2019-07-23
New kernel config file checker

luislinaro:
Add new test to check the kernel config file for the presence of multiple
CONFIG_* values.

* kernel-config-checker.sh: New kernel config file checker script.
* kernel-config-checker.yaml: New kernel config file checker test.

Signed-off-by: Luis Machado <luis.machado@linaro.org>

### DIFF
--- a/automated/linux/kernel-config-checker/kernel-config-checker.sh
+++ b/automated/linux/kernel-config-checker/kernel-config-checker.sh
@@ -1,0 +1,85 @@
+#!/bin/sh
+#
+# Kernel config checker
+#
+# Checks if a set of CONFIG_* values are defined in the config file for a
+# particular version of the Linux kernel.
+#
+# Copyright (C) 2019, Linaro Limited.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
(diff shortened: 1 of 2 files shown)


luislinaro:
This is a variation of an internal test that checks for kernel config parameters. I thought it would be useful to have it upstreamed as opposed to keeping it internal only.

mwasilew:
LGTM. @danrue, @nareshkamboju, @roxell will it be useful for you?

danrue:
Hi @luislinaro - thanks for the test!

Would it be useful to run this for LKFT? It looks like it just checks for duplicate configs, but I'm not sure under what circumstances that would happen. What did you use this test for?

luislinaro:
Dan,

The test is actually supposed to check if particular CONFIG_* values are enabled (compiled in or as a module) in a given kernel config file.

Maybe my explanation of what the test is supposed to achieve wasn't great. Where did you get the duplicates perspective from?

danrue:
Oh, I'm sorry I read too quickly. I misinterpreted "check the kernel config file for the presence of multiple
CONFIG_* values" as duplicates rather than the ability to check for multiple inputs. I understand it now, thank you for the contribution!

luislinaro:
No problem. There was at least one other person that had that same duplicates perspective, so i wondered why. :-)

This should be useful for LKFT if you have a case where you need to validate the presence of a list of CONFIG_* values that are required for the device to work properly.

PR #89, opened 2019-07-23
Kernel config checker

roxell:
LKFTs config file are located /proc/config.gz added support for checking that.


mwasilew:
@roxell 
```
 In automated/linux/kernel-config-checker/kernel-config-checker.sh line 59:
         zcat /proc/config.gz > ${tmpfile}
```

PR #91, opened 2019-07-30
New sysfs bus iio smoke test

luislinaro:
The sysfs bus iio smoke test checks if specific sysfs-bus-iio
properties are within sane bounds, like, for example, temperature or voltage
data.

These tests can be augmented in the future to include string-based property values as opposed to numberic values.

This is a test that was available internally that i've revamped and wanted to make available upstream in case more people find it useful.

Signed-off-by: Luis Machado <luis.machado@linaro.org>

### DIFF
--- a/automated/linux/sysfs-bus-iio-smoke/sysfs-bus-iio-smoke.sh
+++ b/automated/linux/sysfs-bus-iio-smoke/sysfs-bus-iio-smoke.sh
@@ -1,0 +1,91 @@
+#!/bin/sh
+#
+# sysfs bus iio subsystem smoke tests
+#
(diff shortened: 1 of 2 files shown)


mwasilew:
@luislinaro thanks for the patch. Do you mind adding an example usage?

luislinaro:
A practical one or just one example of a valid invocation of the test?

mwasilew:
I'd say one that works on real HW (db410c for example).

luislinaro:
I did not find a working db410c or db820c with mezzanines providing sensor data within reach. I think i might adjust the device list. This test is pretty much generic anyway. As long as someone points the test to an existing file containing sensors data, it should be fine.

PR #93, opened 2019-07-31
[WIP] Add background cmd

igaw:
[This PR is based on add-new-rt-tests PR [#92])

As discussed on IRC here is my attempt to add a background command to the rt-tests.

My first idea to use just environment variable didn't really work out because test shell script don't run call standard hooks. So that's why I added to all tests the background-process-start/stop call.

Also the idea to call lava-background-process-start if available turned out to be a bit a hassle. Because we still need to track state if lava-background-process-start has been called. It looked pretty ugly so I decided to just replicate the whole code in lib.

I think the whole approach looks a bit bolted on. Though I don't have a clear idea how to solve it cleaner except adding some sort of standard hooks (which can be extended or test code injected) which are always called from all tests.

### DIFF
--- a/automated/lib/sh-test-lib
+++ b/automated/lib/sh-test-lib
@@ -7,7 +7,7 @@ error_fatal() {
     # shellcheck disable=SC2039
     local msg="$1"
     [ -z "${msg}" ] && msg="Unknown error"
-    if which lava-test-raise;then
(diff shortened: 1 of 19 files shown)


igaw:
See new PR https://github.com/Linaro/test-definitions/pull/201

PR #95, opened 2019-08-07
Kselftest skiplist

roxell:
rework the skiplist to work with the new run_kselftest.sh


nareshkamboju:
Since we are losing kselftests results without this patch.
now this change is urgent so am going to merge.

PR #99, opened 2019-08-14
kselftests: x86 instead of x86_64 for skipgen

nareshkamboju:


danrue:
I don't understand this one. Where does "x86" come from?

PR #101, opened 2019-08-19
kselftest: lkft: add xfrm_policy.sh to skipfile

roxell:
net: xfrm_policy.sh hangs on i386 running on next

https://bugs.linaro.org/show_bug.cgi?id=5342


nareshkamboju:
This test case hangs intermittently so skipping test.

PR #103, opened 2019-08-21
android boottime: use wait_boot_completed instead of wait_homescreen

liuyq:
as the launcher activity log is not printed in the logcat any more

Change-Id: Ia757ca5c9e9a7a63023e6ba6d55e3e6534d24b88

### DIFF
--- a/automated/android/boottime/boottime.sh
+++ b/automated/android/boottime/boottime.sh
@@ -31,8 +31,8 @@ install_deps 'curl tar xz-utils usbutils' "${SKIP_INSTALL}"
 
 initialize_adb
 adb_root
-# wait till the launcher displayed
-wait_homescreen "${BOOT_TIMEOUT}"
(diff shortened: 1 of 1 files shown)


mwasilew:
IIRC boot completed happens before home screen, so this might lead to some unexpected issues. I have no problems merging this patch, just mentioning this in case you didn't try.

liuyq:
> IIRC boot completed happens before home screen, so this might lead to some unexpected issues. I have no problems merging this patch, just mentioning this in case you didn't try.

Yeah, I know that, but I have not found how to know when the home screen is displayed from the log with the latest P tag and aosp master yet.
Before that, as the surfaceflinger boottime is s major part of the boottime, we still could use that to check the boot time measurement for the android boot part.(another part is the kernel boot part),
so please help to merge this change first.

will update this script when found how to know when the homescreen is displayed.

PR #105, opened 2019-09-04
Add SPDX compliance files

mwasilew:


mwasilew:
@terceiro  could you please check if the copyright part looks right? I think I got licenses correct.

COPYRIGHTS
```diff
@@ -0,0 +1,26 @@
...
+
+Files: *
+Copyright: 2012-2019 Linaro Limited
+           2019 Daniel Wagner <wagi@monom.org>
+           2019 Daniel Wagner <daniel.wagner@siemens.com>
+           2019 Patryk Mungai <patryk.mungai-ndungu.kx@renesas.com>
+           2018 Karsten Tausche <karsten@fairphone.com>
+           2018 Oleksandr Terentiev <oterenti@cisco.com>
+           2017 Lei Yang <Lei.Yang@windriver.com>
+License: GPL-2.0-only
```
terceiro: With this many copyright holders, I would expect one paragraph for each. But this surely works.

terceiro:
LGTM

COPYRIGHTS
mwasilew: I checked all contributors from git log. Is this a correct setup? I can surely split that into separate entries with paths. Some are easy as the contribution was adding new tests. Some cases are more tricky as they were changing a few existing files. Any advice on that?

terceiro: nah I think this should be ok.

mwasilew:
OK, I'll update the list. As I understand lmbench limitation it's about the results not the benchmark itself. We're not distributing the results anyway.

mwasilew:
> Found some missing entries:
> 
> `automated/linux/stream/bin/LICENSE.txt` - not sure what license is that
> 
> `automated/linux/blogbench/bin/COPYING` - ditto

this looks close to 2-clause BSD license
https://opensource.org/licenses/BSD-2-Clause

> 
> `automated/linux/lmbench/bin/COPYING-2` - IMO this actually makes lmbench non-free:
> 
> ```
> 1. You may not distribute results in any public forum, in any publication,       
>    or in any other way if you have modified the benchmarks.                      
>                                                                                  
> 2. You may not distribute the results for a fee of any kind.  This includes      
>    web sites which generate revenue from advertising.   
> ```

this is about the results. Since this repository doesn't produce or distribute any results I think we're OK just using GPLv2 in this case.

I think I will try to separate all binaries into a separate git lfs. Not sure if LAVA supports that though :(

mwasilew:
@terceiro SPDX allows to use licenses that don't exist in their list. I updated the PR.

PR #108, opened 2019-09-10
meminfo: add new test

omnium21:
Add a new test based on automated/android/meminfo,
but remove the Android specific parts.

### DIFF
--- a/automated/linux/meminfo/meminfo.sh
+++ b/automated/linux/meminfo/meminfo.sh
@@ -1,0 +1,30 @@
+#!/bin/sh -e
+# shellcheck disable=SC1091
+
+OUTPUT="$(pwd)/output"
+RESULT_FILE="${OUTPUT}/result.txt"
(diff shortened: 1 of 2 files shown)


omnium21:
Here's a test job where I ran this test:
https://validation.linaro.org/scheduler/job/1933736

mwasilew:
Shellcheck complains about the 'cat':
```
In automated/linux/meminfo/meminfo.sh line 13:
cat /proc/meminfo 2>&1 | tee "${OUTPUT}/proc-meminfo"
     ^-----------^ SC2002: Useless cat. Consider 'cmd < file | ..' or 'cmd file | ..' instead.
```

mwasilew:
I guess in this case shellcheck is wrong. Could you add # shellcheck disable=SC2002 before the line with cat?

omnium21:
Thanks Milosz. I added the change and rebased to master, since the docker change was merged already.

PR #109, opened 2019-09-10
docker: hello-world support

omnium21:
Add support for running the "hello-world" sample image to the docker
test.

This image requires fewer resources than an ubuntu system running a
shell command.

To run hello-world, the caller passes in "hello-world" for the IMAGE
parameter.


omnium21:
Here's a test job where I ran this test:
https://validation.linaro.org/scheduler/job/1933736

PR #119, opened 2019-10-18
automated/linux: classify UNRESOLVED as fail

mwasilew:
Open Posix test suite may report test result as UNRESOLVED. This might
mean ABI break and should be classified as FAIL in test-definitions
convention.


nareshkamboju:
This patch might create some level of noise on results.
But it is ok.

PR #121, opened 2019-10-24
automated/linux: Use tags for 2.x in workload-automation

alimon:
We have a workload-automation3 testdef so this one is target to use
the version 2 of the tool use proper tags by default.

Change-Id: I9121a14772fe59d750c68d97c59fb2f3ab481004
Signed-off-by: Aníbal Limón <anibal.limon@linaro.org>


mwasilew:
At some point we should probably abandon v2.

PR #124, opened 2019-10-27
Improve accuracy and completeness of results 

paolo-github:


automated/linux/ssuite/run-bench.sh
```diff
@@ -181,10 +181,10 @@ run_test() {
 	# measured is reported. For each passed test case. Here is the
 	# format of possible lines in the result file:
 	#
-	# throughput-<scheduler1_name>--<workload_name> pass <real number> MB/s
+	# throughput-<scheduler1_name>--<workload_name> pass <min> <max> <avg> <std dev> MB/s
```
mwasilew: Firstly this is incorrect in terms of 'send-to-lava.sh' which expects just 1 measurement per result (LAVA only accepts one measurement). Secondly if you look at lines 191-194 the numbers you refer to are already recorded. I propose to drop this commit and only merge the increase in number of repetitions. It would be even better to use variable for number of loops so it can be adjusted in the LAVA job definition.

PR #125, opened 2019-11-12
RFC: Add uart-loopback test with YepKit support

omnium21:
Add a test to perform loopback tests across two UARTs.

This test is a wrapper around an existing Schneider Electric UART
loopback test, with minor mods to their test for output formatting.

The test integrates ykush/YepKit support, used to enable/disable the
FTDI USB UART device. It will be broken out somehow in another revision
of the test.


omnium21:
This is still work-in-progress, but I wanted to share an early version that works.

Currently, the Yepkit support is built into the uart-loopback test, which needs to change. That's the main bit I'm interested in initially: what to do with the YepKit support?

- Make it a library in the sh-test-lib, or somewhere similar?
- Make it a separate test that the job definition calls before and after running a test that needs to manipulate the Yepkit ports?
- Or something else?

omnium21:
Here's a LAVA test job where I ran this:
https://validation.linaro.org/scheduler/job/1948000

mwasilew:
Copying output of sanity check
```
* SHELLCHECK: [FAILED]: automated/linux/uart-loopback/test-uart-loopback.sh
* SHELLCHECK: [OUTPUT]:

 In automated/linux/uart-loopback/test-uart-loopback.sh line 23:
 device_exists ${TXUART}
               ^-------^ SC2086: Double quote to prevent globbing and word splitting.

 In automated/linux/uart-loopback/test-uart-loopback.sh line 24:
 device_exists ${RXUART}
               ^-------^ SC2086: Double quote to prevent globbing and word splitting.

 In automated/linux/uart-loopback/test-uart-loopback.sh line 32:
 	${DIR}/uart-loopback -o ${TXUART} -i ${RXUART} -s ${LENGTH} -r
         ^----^ SC2086: Double quote to prevent globbing and word splitting.
                                 ^-------^ SC2086: Double quote to prevent globbing and word splitting.
                                              ^-------^ SC2086: Double quote to prevent globbing and word splitting.
                                                           ^-------^ SC2086: Double quote to prevent globbing and word splitting.


 In automated/linux/uart-loopback/test-uart-loopback.sh line 33:
 	if [ "$?" -ne "0" ]; then ERRORS=$((ERRORS+1)); fi
              ^--^ SC2181: Check exit code directly with e.g. 'if mycmd;', not indirectly with $?.
 
 In automated/linux/uart-loopback/test-uart-loopback.sh line 39:
 	stty -F ${TXUART} ${SETTINGS}
                 ^-------^ SC2086: Double quote to prevent globbing and word splitting.
                           ^---------^ SC2086: Double quote to prevent globbing and word splitting.

 In automated/linux/uart-loopback/test-uart-loopback.sh line 40:
 	stty -F ${RXUART} ${SETTINGS}
                 ^-------^ SC2086: Double quote to prevent globbing and word splitting.
                           ^---------^ SC2086: Double quote to prevent globbing and word splitting.
 
 In automated/linux/uart-loopback/test-uart-loopback.sh line 44:
 	for length in `seq 1 33`; do
                       ^--------^ SC2006: Use $(...) notation instead of legacy backticked `...`.
 
 In automated/linux/uart-loopback/test-uart-loopback.sh line 45:
 		test_one ${length}
                          ^-------^ SC2086: Double quote to prevent globbing and word splitting.
 
 In automated/linux/uart-loopback/test-uart-loopback.sh line 61:
 for baud in ${baudrates[@]} ;
             ^-------------^ SC2068: Double quote array expansions to avoid re-splitting elements.

 For more information:
   https://www.shellcheck.net/wiki/SC2068 -- Double quote array expansions to ...
   https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...
   https://www.shellcheck.net/wiki/SC2006 -- Use $(...) notation instead of le...
* UNKNOWN [SKIPPED]: automated/linux/uart-loopback/uart-loopback.c - Unknown file type detected: text/x-c
* SHELLCHECK: [FAILED]: automated/linux/uart-loopback/uart-loopback.sh
* SHELLCHECK: [OUTPUT]:

 In automated/linux/uart-loopback/uart-loopback.sh line 9:
 TEST_SUITE="uart-loopback"
 ^--------^ SC2034: TEST_SUITE appears unused. Verify use (or export if used externally).

 In automated/linux/uart-loopback/uart-loopback.sh line 105:
 				pushd ykush 2>&1 > /dev/null
                                 ^--------------------------^ SC2164: Use 'cd ... || exit' or 'cd ... || return' in case cd fails.
                                             ^--^ SC2069: To redirect stdout+stderr, 2>&1 must be last (or use '{ cmd > file; } 2>&1' to clarify).

 In automated/linux/uart-loopback/uart-loopback.sh line 109:
 				popd 2>&1 > /dev/null
                                 ^-------------------^ SC2164: Use 'cd ... || exit' or 'cd ... || return' in case cd fails.
                                      ^--^ SC2069: To redirect stdout+stderr, 2>&1 must be last (or use '{ cmd > file; } 2>&1' to clarify).

 
 In automated/linux/uart-loopback/uart-loopback.sh line 135:
 		${YKUSHCMD} -${updown} ${YKUSH_PORT} 2>&1 > /dev/null
                                        ^-----------^ SC2086: Double quote to prevent globbing and word splitting.
                                                      ^--^ SC2069: To redirect stdout+stderr, 2>&1 must be last (or use '{ cmd > file; } 2>&1' to clarify).

 In automated/linux/uart-loopback/uart-loopback.sh line 154:
         for ((i=0;i<${retries};++i)); do
                     ^--------^ SC2004: $/${} is unnecessary on arithmetic variables.

 In automated/linux/uart-loopback/uart-loopback.sh line 155:
 		local exists=$(device_exists ${device})
                       ^----^ SC2155: Declare and assign separately to avoid masking return values.
                                              ^-------^ SC2086: Double quote to prevent globbing and word splitting.

 In automated/linux/uart-loopback/uart-loopback.sh line 168:
 	local logfile=$(mktemp "/tmp/$(basename $0).XXXXXXXXXXXX")
               ^-----^ SC2155: Declare and assign separately to avoid masking return values.
                                                 ^-- SC2086: Double quote to prevent globbing and word splitting.

 In automated/linux/uart-loopback/uart-loopback.sh line 170:
 	echo Testing data transfer from ${UART0} to ${UART1} | tee ${logfile}
                                         ^------^ SC2086: Double quote to prevent globbing and word splitting.
                                                     ^------^ SC2086: Double quote to prevent globbing and word splitting.

 In automated/linux/uart-loopback/uart-loopback.sh line 171:
 	./test-uart-loopback.sh ${UART0} ${UART1} | tee -a ${logfile}
                                 ^------^ SC2086: Double quote to prevent globbing and word splitting.
                                          ^------^ SC2086: Double quote to prevent globbing and word splitting.

 In automated/linux/uart-loopback/uart-loopback.sh line 172:
 	sed -i 's/\.//g' ${logfile}

 In automated/linux/uart-loopback/uart-loopback.sh line 173:
 	grep -e "pass" -e "fail" ${logfile} >> ${RESULT_FILE}
                                                ^------------^ SC2086: Double quote to prevent globbing and word splitting.

 In automated/linux/uart-loopback/uart-loopback.sh line 185:
 ykush_port ${YKUSH_PORT} on
            ^-----------^ SC2086: Double quote to prevent globbing and word splitting.

 In automated/linux/uart-loopback/uart-loopback.sh line 186:
 wait_for_device ${UART0}
                 ^------^ SC2086: Double quote to prevent globbing and word splitting.

 In automated/linux/uart-loopback/uart-loopback.sh line 187:
 wait_for_device ${UART1}
                 ^------^ SC2086: Double quote to prevent globbing and word splitting.

 In automated/linux/uart-loopback/uart-loopback.sh line 189:
 ykush_port ${YKUSH_PORT} off
            ^-----------^ SC2086: Double quote to prevent globbing and word splitting.

 For more information:
   https://www.shellcheck.net/wiki/SC2034 -- TEST_SUITE appears unused. Verify...
   https://www.shellcheck.net/wiki/SC2069 -- To redirect stdout+stderr, 2>&1 m...
   https://www.shellcheck.net/wiki/SC2155 -- Declare and assign separately to ..
```

omnium21:
I fixed most of the sanity-check.sh errors, but these two remain, and I don't know how to fix them:

In automated/linux/uart-loopback/test-uart-loopback.sh line 43:
 	stty -F "${TXUART}" ${settings}
                             ^---------^ SC2086: Double quote to prevent globbing and word splitting.
 
 
 In automated/linux/uart-loopback/test-uart-loopback.sh line 44:
 	stty -F "${RXUART}" ${settings}
                             ^---------^ SC2086: Double quote to prevent globbing and word splitting.

Quoting the variable causes the stty command to fail. stty seems to *want* word splitting to take place.

mwasilew:
@omnium21 if there is a legitimate reason for not following shellcheck, please add a comment explaining the problem and in the line preceding the warning add:
```
# shellcheck disable=SC2086
```

automated/linux/uart-loopback/uart-loopback.yaml
```diff
@@ -0,0 +1,32 @@
...
+        - uart
+    devices:
+        - rzn1d
+        - soca9
+
+params:
+        UART0: "/dev/ttyUSB0"
+        UART1: "/dev/ttyS1"
+        YKUSHPORT: 1
+        SKIP_INSTALL: "True"
```
roxell: I think that SKIP_INSTALL normally means that we skip to install the tests and its dependencies.

automated/linux/uart-loopback/uart-loopback.yaml
```diff
@@ -0,0 +1,32 @@
+# SPDX-License-Identifier: GPL-2.0
+# Copyright (C) 2019 Schneider Electric
+metadata:
+    format: Lava-Test Test Definition 1.0
+    name: UART Loopback
+    description: "...
+                     ..."
```
roxell: Should this be filled in?

omnium21:
> Would it make sense to put the c-file and Makefile on a github repository so that can be updated and worked on separate from the test-definitions' files ?
> 
> I'm not saying its wrong to do 'make' here, if this repositories only contains uart-looback.yaml and uart-loopback.sh and the other files lives in another repository. That would mean that someone the uses OE would be able to write a recipe to install the uart-looback into their OE rootfs.

That is possible, for sure. I copied those files from an internal private repo to here to make them public, and keep them in one place. But yes, I could put them into a different repo without too much effort.

Rather than creating my own, I wonder if it would make sense to have a Linaro QA code repo? Something a bit like this repo, but for various bits of code.

automated/linux/uart-loopback/uart-loopback.yaml
omnium21: Yes, good point :)

omnium21:
First: apologies for polluting the branch in this PR with other commit. I'd intended using a branch specifically for the PR, but ended up using it for other random test work. I'll sort something out when I push an improved version of the uart-loopback test.

Meanwhile, I've been thinking about the issue with integrating Yepkit support into the test. Right now, the test knows about Yepkit, builds it if needed, enables the port, runs the test, then disables the port.

I think I should split the Yepkit stuff out from the uart-loopback test, and the test should not know about it at all. Yepkit could be a separate "test" that is run as a distinct step, eg, in LAVA yaml:

- name: yepkit
  from: git
  repository: https://...
  path: ..../yepkit.yaml
  parameters:
    ENABLE_PORTS: '1'

- name: uart-loopback
      from: git
      history: false
      repository: https://github.com/Linaro/test-definitions.git
      branch: linaro
      path: automated/linux/uart-loopback/uart-loopback.yaml
      parameters:
        SKIP_INSTALL: 'true'
        UART0: /dev/ttyUSB0
        UART1: /dev/ttyS1

- name: yepkit
  from: git
  repository: https://...
  path: ..../yepkit.yaml
  parameters:
    DISABLE_PORTS: '1'

The additional yepkit.yaml could take parameters with lists of ports to enable and/or disable, and have the logic to install and/or build the yepkit library if it isn't present in the rootfs.

roxell:
> > Would it make sense to put the c-file and Makefile on a github repository so that can be updated and worked on separate from the test-definitions' files ?
> > I'm not saying its wrong to do 'make' here, if this repositories only contains uart-looback.yaml and uart-loopback.sh and the other files lives in another repository. That would mean that someone the uses OE would be able to write a recipe to install the uart-looback into their OE rootfs.
> 
> That is possible, for sure. I copied those files from an internal private repo to here to make them public, and keep them in one place. But yes, I could put them into a different repo without too much effort.

I think that would be great.

> 
> Rather than creating my own, I wonder if it would make sense to have a Linaro QA code repo? Something a bit like this repo, but for various bits of code.

I'm not sure about that, @mwasilew  do you know if we have a repository for that?
or maybe just create a new repository under github.com/Linaro/ ?

roxell:
> First: apologies for polluting the branch in this PR with other commit. I'd intended using a branch specifically for the PR, but ended up using it for other random test work. I'll sort something out when I push an improved version of the uart-loopback test.
> 
> Meanwhile, I've been thinking about the issue with integrating Yepkit support into the test. Right now, the test knows about Yepkit, builds it if needed, enables the port, runs the test, then disables the port.
> 
> I think I should split the Yepkit stuff out from the uart-loopback test, and the test should not know about it at all. Yepkit could be a separate "test" that is run as a distinct step, eg, in LAVA yaml:
> 
> * name: yepkit
>   from: git
>   repository: https://...
>   path: ..../yepkit.yaml
>   parameters:
>   ENABLE_PORTS: '1'
> * name: uart-loopback
>   from: git
>   history: false
>   repository: https://github.com/Linaro/test-definitions.git
>   branch: linaro
>   path: automated/linux/uart-loopback/uart-loopback.yaml
>   parameters:
>   SKIP_INSTALL: 'true'
>   UART0: /dev/ttyUSB0
>   UART1: /dev/ttyS1
> * name: yepkit
>   from: git
>   repository: https://...
>   path: ..../yepkit.yaml
>   parameters:
>   DISABLE_PORTS: '1'
> 
> The additional yepkit.yaml could take parameters with lists of ports to enable and/or disable, and have the logic to install and/or build the yepkit library if it isn't present in the rootfs.

I like that idea.

omnium21:
I've stripped out the Yepkit stuff and removed the unrelated patches. I'll submit all that as a separate PR.

This doesn't actually work any more, so apart from the Travis checks failing, I need to fix it...

It's this or a related change that breaks it:
-       stty -F "${TXUART}" ${settings}
-       stty -F "${RXUART}" ${settings}
+       stty -F "${TXUART}" "${settings}"
+       stty -F "${RXUART}" "${settings}"

I get errors like this when I run the script:
stty: invalid argument '9600 -parenb -cstopb -crtscts cs8 -ignbrk -brkint  -parmrk -istrip -inlcr -igncr -icrnl -ixon -opost -echo -echonl -icanon -isig -iexten'

omnium21:
Basically, the fix for this travis error - adding quotes - doesn't work:
https://travis-ci.com/github/Linaro/test-definitions/jobs/302904496#L1267

mwasilew:
@omnium21 just add
```
# shellcheck disable=SC2086
```
before the line which requires no quotes.

omnium21:
Just realised I hadn't added "functional" to the scope field, and "loopback" while I was there.

automated/linux/uart-loopback/uart-loopback.yaml
mwasilew: +1 this is the convention with LTP for example. So this means that build_loopback should be placed inside install(). As a side effect Makefile and uart_loopback.c should be moved outside of this PR (and possibly repository. 

Alternative solution is providing pre-built binaries and storing them in the repository. This is done in case of some benchmarks. It requires to have at least x86 and arm versions of binaries. IMHO this option is simpler and I would prefer it.

omnium21:
I've updated the commit to move the build inside "install()".
I can provide x86 and arm binaries too, if you like. Not sure how I decide which binary to run. Something with uname, I guess?

mwasilew:
There is 'detect_abi' in sh-test-lib: https://github.com/Linaro/test-definitions/blob/master/automated/lib/sh-test-lib#L183

omnium21:
Yes, thanks, I've used detect_abi in the latest version.
I made quite a few changes:

- I removed uart-loopback.c and Makefile as requested
- I added 3 x binaries for armeabi, arm64 and x86_64
- I combined test-uart-loopback.sh into uart-loopback.sh
- I removed all the install parts as nothing is needed now
- I renamed the parameters to -t (txuart) and -r (rxuart)

Maybe some others too...

omnium21:
Oh, and I ran it through LAVA
https://validation.linaro.org/scheduler/job/1979318

mwasilew:
This looks good. There are 2 more shellcheck warnings to fix:
```
* SHELLCHECK: [FAILED]: ./automated/linux/uart-loopback/uart-loopback.sh
* SHELLCHECK: [OUTPUT]:
 In ./automated/linux/uart-loopback/uart-loopback.sh line 10:
 SKIP_INSTALL="false"
 ^----------^ SC2034: SKIP_INSTALL appears unused. Verify use (or export if used externally).

 In ./automated/linux/uart-loopback/uart-loopback.sh line 91:
 	case "$abi" in
               ^--^ SC2154: abi is referenced but not assigned.

 For more information:
   https://www.shellcheck.net/wiki/SC2034 -- SKIP_INSTALL appears unused. Veri...
   https://www.shellcheck.net/wiki/SC2154 -- abi is referenced but not assigned.
```

omnium21:
OK, I fixed those. The SKIP_INSTALL was a simple omission, and I shellcheck disabled the $abi warning.

PR #135, opened 2020-01-30
automated: linux. kvm-unit-tests: check if GIT_REF

roxell:


nareshkamboju:
The idea is to test known good version else exit as fail.
This is perfect way.

PR #138, opened 2020-02-13
igt: power cycle Chamelium once it is unreachable

arthur-she:
1. Power cycle the Chamelium once we don't get the ping response from it
    (snmpset from Net-SNMP is required)
2. Check uboot console lock status and uptime of the Chamelium
3. Delete chameliumd log after each test


arthur-she:
Test job:
[The target board tries to reboot the Chamelium to get it back to work.](https://validation.linaro.org/scheduler/job/1967595#L1859)

PR #144, opened 2020-03-04
files-exist test added

omnium21:
Often, it is useful to check if files exist on the board, for example,
before running certain tests, to make sure that the driver is
detecting hardware, or to check the kernel is configured
correctly.

No effort is made to ensure that a file is a character, network or block
device.

In the SE case, we want to make sure that certain device nodes exist,
for example. This test contains knowledge of SE specific files and
device nodes across boards and kernel versions.

The caller can augment the list of files by passing in the '-e <extra
files> string with a list of additional files.


mwasilew:
this is similar to bootrr (https://github.com/andersson/bootrr). Could you check if bootrr can be used instead?

omnium21:
I'll implement this in bootrr, if possible, thanks for the recommendation.

PR #145, opened 2020-03-04
iperf: split into client and server

omnium21:
Split the iperf.sh script so it runs as either server of client, but not
both.

If the caller wishes to run the test with a local server, they must
invoke the script twice.

This commit also creates a YAML for the server.

Additional parameters are added to the client to allow CPU affinity, or
to reverse the test, so the client is the receiver instead of the
transmitter.

### DIFF
--- a/automated/linux/iperf/iperf-server.yaml
+++ b/automated/linux/iperf/iperf-server.yaml
@@ -1,0 +1,41 @@
+metadata:
+    name: iperf
+    format: "Lava-Test-Shell Test Definition 1.0"
+    description: "iperf is a tool for active measurements of the maximum
+                  achievable bandwidth on IP networks."
+    maintainer:
+        - milosz.wasilewski@linaro.org
+    os:
+        - debian
+        - ubuntu
+        - fedora
+        - centos
+        - openembedded
+    scope:
+        - performance
+    environment:
+        - lava-test-shell
+    devices:
+        - hi6220-hikey
+        - apq8016-sbc
+        - mustang
+        - moonshot
+        - thunderX
+        - d03
+        - d05
+        - soca9
+        - rzn1d
+
+params:
+    SKIP_INSTALL: "false"
+    SERVER_ETHERNET_DEVICE: "eth0"
+
+run:
+    steps:
+        - cd ./automated/linux/iperf/
+        - ipaddr=$(lava-echo-ipv4 ${SERVER_ETHERNET_DEVICE} | tr -d '\0')
+        - if [ -z "${ipaddr}" ]; then echo lava-test-raise "${SERVER_ETHERNET_DEVICE} not found"; fi
+        - lava-send server-ready ipaddr=${ipaddr}
+        - ./iperf.sh -s "${SKIP_INSTALL}"
+        - lava-wait client-done
+        - ../../utils/send-to-lava.sh ./output/result.txt
--- a/automated/linux/iperf/iperf.sh
+++ b/automated/linux/iperf/iperf.sh
@@ -5,27 +5,34 @@
 OUTPUT="$(pwd)/output"
 RESULT_FILE="${OUTPUT}/result.txt"
 LOGFILE="${OUTPUT}/iperf.txt"
-# Test localhost by default, which tests the efficiency of TCP/IP stack.
-# To test physical network bandwidth, specify remote test server with '-c'.
-# Execute 'iperf3 -s' on remote host to run iperf3 test server.
-SERVER="127.0.0.1"
+# If SERVER is blank, we are the server, otherwise
+# If we are the client, we set SERVER to the ipaddr of the server
+SERVER=""
 # Time in seconds to transmit for
 TIME="10"
 # Number of parallel client streams to run
 THREADS="1"
 # Specify iperf3 version for CentOS.
 VERSION="3.1.4"
+# By default, the client sends to the server,
+# Setting REVERSE="-R" means the server sends to the client
+REVERSE=""
+# CPU affinity is blank by default, meaning no affinity.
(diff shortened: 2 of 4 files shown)


omnium21:
https://validation.linaro.org/scheduler/job/1971967.1#L943

The metrics seem to be working now too. I see these two lines in the job:

```
iperf_sender pass 94.1 Mbits/sec
iperf_receiver pass 94.1 Mbits/sec
```

And this translates to:

```
case: iperf_sender
case_id: 131821606
definition: 0_iperf3-test
measurement: 94.1
result: pass
units: Mbits/sec

case: iperf_receiver
case_id: 131821607
definition: 0_iperf3-test
measurement: 94.1
result: pass
units: Mbits/sec
```

I think they weren't working because the test case/name had a dash in it. So I've replaced the dash with an underscore.

omnium21:
I have a few small changes yet to push...

mwasilew:
It worked here: https://validation.linaro.org/results/1971869/0_iperf3-test with dashes

automated/linux/iperf/iperf-server.yaml
```diff
@@ -0,0 +1,38 @@
+metadata:
+    name: iperf
+    format: "Lava-Test-Shell Test Definition 1.0"
+    description: "iperf is a tool for active measurements of the maximum
+                  achievable bandwidth on IP networks."
+    maintainer:
+        - chase.qi@linaro.org
```
mwasilew: please change maintainer to my email.

automated/linux/iperf/iperf-server.yaml
```diff
@@ -0,0 +1,38 @@
...
+        - rzn1d
+
+params:
+    SKIP_INSTALL: "false"
+    SERVER_ETHERNET_DEVICE: "eth0"
+
+run:
+    steps:
+        - cd ./automated/linux/iperf/
+        - lava-send server-ready ipaddr=$(lava-echo-ipv4 ${SERVER_ETHERNET_DEVICE})
```
mwasilew: maybe this should be split into to lines. In case lava-echo-ipv4 returns empty string we should abort with lava-test-raise.

automated/linux/iperf/iperf.sh
```diff
@@ -15,17 +15,23 @@ TIME="10"
...
+AFFINITY="-A 0"
 
 usage() {
-    echo "Usage: $0 [-c server] [-t time] [-p number] [-v version] [-s true|false]" 1>&2
+    echo "Usage: $0 [-c server] [ -e server eth device] [-t time] [-p number] [-v version] [-R] [-s true|false]" 1>&2
     exit 1
 }
 
-while getopts "c:t:p:v:s:h" o; do
+while getopts "A:c:t:p:v:s:Rh" o; do
```
mwasilew: -e is missing

mwasilew:
I'm a bit confused how iperf.sh runs both server and client. I didn't see client code in the script. Did I miss it or it's coming in the next update?

automated/linux/iperf/iperf.sh
omnium21: I dropped -e and didn't update correctly. I realised iperf-server.yaml is the only one that needs to know the server's ethernet address, since we keep the lava-* commands in the yaml only.

omnium21:
The client code was always in iperf.sh. I've added the server code to it, which is small: just the "iperf3 -s -D" call. But I need to add a test result for that step. I'll add some comments.

Let me push some more changes and hopefully it's clearer.

automated/linux/iperf/iperf-server.yaml
omnium21: Ack.

omnium21:
> It worked here: https://validation.linaro.org/results/1971869/0_iperf3-test with dashes

Yes, it was my misunderstanding. When the test runs, the metrics are shown in the live log. But when you reload the page after the job completes, the metrics appear.

automated/linux/iperf/iperf-server.yaml
omnium21: I missed this comment, I'll address it now...

omnium21:
Pah! I'm *still* not finished... just noticed a few other minor tweaks I can make...

automated/linux/iperf/iperf-server.yaml
omnium21: Ugh, I split it and get this error:
/lava-1972096/0/tests/0_iperf3-test/run.sh: line 24: unexpected EOF while looking for matching `\"'
I guess I need to run it through "tr" or something to deal with the null.

automated/linux/iperf/iperf-server.yaml
```diff
@@ -0,0 +1,40 @@
...
+        - rzn1d
+
+params:
+    SKIP_INSTALL: "false"
+    SERVER_ETHERNET_DEVICE: "eth0"
+
+run:
+    steps:
+        - cd ./automated/linux/iperf/
+        - ipaddr=$(lava-echo-ipv4 ${SERVER_ETHERNET_DEVICE} | tr -d '\0')
```
mwasilew: the error is printed to stderr so it's not harmful. I would add one more line after this:
```
if [ -z "${ipaddr}" ]; then echo lava-test-raise "${SERVER_ETHERNET_DEVICE} not found"; fi
```
This will abort the test in case the interface doesn't exist. IHMO this is right thing to do.

omnium21: Ack.

omnium21: OK, done in the latest push.

omnium21:
My latest job with this version:
https://validation.linaro.org/scheduler/job/1972181

PR #149, opened 2020-03-09
iperf: move lava commands into iperf.sh

omnium21:
Rather than run the LAVA commands in the YAML, run them in the script.

This is particularly useful so that installation happens before the lava-wait.
This helps prevent timeouts on slower platforms.

### DIFF
--- a/automated/linux/iperf/iperf.sh
+++ b/automated/linux/iperf/iperf.sh
@@ -20,16 +20,18 @@ REVERSE=""
 # CPU affinity is blank by default, meaning no affinity.
 # CPU numbers are zero based, eg AFFINITY="-A 0" for the first CPU
 AFFINITY=""
+ETH="eth0"
 
 usage() {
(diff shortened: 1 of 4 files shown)


automated/linux/iperf/iperf.sh
```diff
@@ -65,6 +67,20 @@ fi
 
 # Run local iperf3 server as a daemon when testing localhost.
 if [ "${SERVER}" = "" ]; then
+    cmd="lava-echo-ipv4"
+    if which "${cmd}"; then
+        ipaddr=$(${cmd} "${ETH}" | tr -d '\0')
+        if [ -z "${ipaddr}" ]; then
+            echo "lava-test-raise ${ETH} not found";
```
mwasilew: if lava-echo-ipv4 is defined this means we're running in lava so lava-test-raise should also be defined. In such case don't use 'echo' but call lava-test-raise.

omnium21: Oh yeah!

PR #158, opened 2020-03-24
Update armnn-mlperf.sh

K1504296:
armnn-ci-build: Adding apt-get update.


automated/linux/armnn-benchmarks/armnn-mlperf.sh
```diff
@@ -26,6 +26,7 @@ done
 ! check_root && error_msg "You need to be root to run this script."
 create_out_dir "${OUTPUT}"
 
+apt-get update
```
mwasilew: this shouldn't be necessary
https://github.com/Linaro/test-definitions/blob/master/automated/lib/sh-test-lib#L243
Maybe system time isn't correct? The proper fix should go to sh-test-lib anyway. Tests should be distro agnostic.

PR #159, opened 2020-03-26
igt: modify test definition for igt_runner

arthur-she:
There is a test runner, igt_runner, comes with the newer version of
igt-gpu-tests. We can run the tests with it rather than piglit. Now the
test definition is compatible with both igt_runner and piglit.


mwasilew:
@arthur-she could you convert the python script to be python3 compatible? This should be a separate PR.

arthur-she:
> @arthur-she could you convert the python script to be python3 compatible? This should be a separate PR.

@mwasilew , I will do it later, thanks for the review.

PR #162, opened 2020-04-09
block-device test added

omnium21:
Test a partition on a block device.

Check that files can be written and read reliably, measure the timings
for read and write, and run the bonnie++ block device test on the
device.

### DIFF
--- a/automated/linux/block-device/block-device.sh
+++ b/automated/linux/block-device/block-device.sh
@@ -1,0 +1,261 @@
+#!/bin/bash
+# SPDX-License-Identifier: GPL-2.0-only
+# Copyright (C) 2019 Linaro Ltd.
+
+# shellcheck disable=SC1091
+. ../../lib/sh-test-lib
+OUTPUT="$(pwd)/output"
+RESULT_FILE="${OUTPUT}/result.txt"
+
+PARTDEV=""
+PARTNAME=""
+DEVICE=""
+DEVNAME=""
+
+SKIP_INSTALL="false"
+FORMAT_DEVICE="false"
+
+MOUNT_DIR=$(mktemp -d "/tmp/block-device.XXXXX")
+TMPFILE=$(mktemp "/tmp/10M.XXXXX")
+RESULT=pass
+
+usage() {
+	echo "\
+	Usage: $0
+		     [-d </dev/sdXN>]
+		     [-f <true|false>]
+		     [-s <true|false>]
+
+	-d </dev/sdXN>
+		This is the block device to be tested, eg. /dev/sda1.
+		It should be a partition, not a raw block device
+	-f <true|false>
+		This will erase the device and create a new partition table
+		with a single ext4 partition on it.
+	-s <true|false>
+		Tell the test to skip installation of dependencies, or not
+
+	This test will perform tests on a block device.
+	"
+}
+
+while getopts "d:f:s:h" opts; do
+	case "$opts" in
+		d) PARTDEV="${OPTARG}";
+		   PARTNAME=$(basename -- "${PARTDEV}");
+		   DEVICE=${PARTDEV/%[[:digit:]]/}
+		   DEVNAME=$(basename -- "${DEVICE}")
+		   ;;
+		f) FORMAT_DEVICE="${OPTARG,,}" ;;
+		s) SKIP_INSTALL="${OPTARG,,}" ;;
+		h|*) usage ; exit 1 ;;
+	esac
+done
+
+install() {
+	if [ "${SKIP_INSTALL}" = "true" ]; then
+		info_msg "Skip installing dependencies"
+	else
+		# install dependencies
+		dist=
+		dist_name
+		case "${dist}" in
+			debian|ubuntu)
+				pkgs="bonnie++"
+				install_deps "${pkgs}" "${SKIP_INSTALL}"
+				;;
+			fedora|centos)
+				pkgs="bonnie++"
+				install_deps "${pkgs}" "${SKIP_INSTALL}"
+				;;
+			# When we don't have a package manager
+			# Assume dependencies pre-installed
+			*)
+				echo "Unsupported distro: ${dist}! Package installation skipped!"
+				;;
+		esac
+	fi
+}
+
+block_device_exists () {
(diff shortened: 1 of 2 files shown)


omnium21:
I've seen the Travis error and reproduce them below for my own convenience when fixing... I'll have to work out how Travis runs sanity-check.sh, because these errors don't happen locally.

* METADATA [FAILED]: ./automated/linux/block-device/block-device.yaml
	'name' contains whitespace
Validating file: ./automated/linux/block-device/block-device.sh
* SHELLCHECK: [FAILED]: ./automated/linux/block-device/block-device.sh
* SHELLCHECK: [OUTPUT]:
 
 In ./automated/linux/block-device/block-device.sh line 46:
 		   DEVICE=$(echo "${PARTDEV}" | sed 's/[0-9]$//g');
                             ^-- SC2001: See if you can use ${variable//search/replace} instead.
 
 
 In ./automated/linux/block-device/block-device.sh line 161:
 			local seconds=$(cat "${timings}" | tail -1 | awk '{print $8}')
                               ^-----^ SC2155: Declare and assign separately to avoid masking return values.
                                             ^----------^ SC2002: Useless cat. Consider 'cmd < file | ..' or 'cmd file | ..' instead.
 
 
 In ./automated/linux/block-device/block-device.sh line 162:
 			local mbps=$(cat "${timings}" | tail -1 | awk '{print $10}')
                               ^--^ SC2155: Declare and assign separately to avoid masking return values.
                                          ^----------^ SC2002: Useless cat. Consider 'cmd < file | ..' or 'cmd file | ..' instead.
 
 
 In ./automated/linux/block-device/block-device.sh line 182:
 	copy_timing_test "write" "${TMPFILE}" "${MOUNT_DIR}/$(basename -- ${TMPFILE})"
 
 
 In ./automated/linux/block-device/block-device.sh line 188:
 	copy_timing_test "read" "${MOUNT_DIR}/$(basename -- ${TMPFILE})" "${TMPFILE}"
                                                             ^--------^ SC2086: Double quote to preven

omnium21:
I shellcheck disabled SC2086 on lines 185 and 192 because it was giving me a recursive "add double quotes" error. And it wanted me to double quote the function name, which seems ridiculous.

mwasilew:
shellcheck version usually makes a difference. Travis just runs './sanity-check.sh', no magic here :)

mwasilew:
We used to have standalone bonnie++ test but it got lost during refactoring. Do you think it makes sense to bring it back?

omnium21:
Having a standalone test might be useful. Depends what it does, I guess. My test is only checking bonnie++ runs successfully and passes all tests, it isn't parsing the subtests. So if you have one that does, it would be better. I've never seen a bonnie++ test fail, so I'm not sure how valuable it is, hence why I didn't bother parsing it.

My PR does two tests: a simple read/write test and the bonnie++ test. I'm note sure if there's extra value in splitting those two out. I'd be happy to add it to the old bonnie++ test instead, of course.

Either way, one thing I found important was the format option. Relying on an unknown USB stick or SDcard hasn't worked out well for me in the past. This would be common code between the read/write test and the bonnie++ test, so if we are to split the tests, I'd like some way to make the format code common. I suppose other tests may want something similar?

mwasilew:
bonnie++ is a benchmark and IIRC the test was just parsing the benchmark results. Let's merge this as it is and I'll try to revive standalone bonnie++. Let me run it locally 1st with test-runner as well. Should be done today.

automated/linux/block-device/block-device.sh
```diff
@@ -0,0 +1,229 @@
...
+}
+
+while getopts "d:fs:h" opts; do
+	case "$opts" in
+		d) PARTDEV="${OPTARG}";
+		   PARTNAME=$(basename -- "${PARTDEV}");
+		   DEVICE=${PARTDEV/%[[:digit:]]/}
+		   DEVNAME=$(basename -- "${DEVICE}")
+		   ;;
+		f) FORMAT_DEVICE="true" ;;
```
mwasilew, line 49: I think this is wrong. At least this isn't in line with the YAML file, where it's always present.

automated/linux/block-device/block-device.sh
```diff
@@ -0,0 +1,229 @@
...
+	if [[ -b ${device} ]]; then
+		echo true
+	else
+		echo false
+	fi
+}
+
+wait_for_device () {
+	local device=$1
+        local retries=20
```
mwasilew, line 92: this indentation uses spaces and all other indents use tabs.

omnium21: Ooops!

automated/linux/block-device/block-device.sh
omnium21: Ah yes, I've started to convert it from "-f value" to just "-f" and not completed the change. But it needs to have the value for the YAML to make sense. I'll fix...

omnium21:
> Here is commit where we had that:
> [d3d1d1d#diff-086662f97164101b02f14da9640448c5](https://github.com/Linaro/test-definitions/commit/d3d1d1de42afb31ee58781d13462d26a5b04bda9#diff-086662f97164101b02f14da9640448c5)

Thanks, I'll integrate that into this test.

> It was than 'lost' in the refactoring as I believe there was no interest in running this test.

I expect there never used to be a need, but I'm reviving it now :)

mwasilew:
@omnium21 I think it's good to be merged. Should I wait for the bonnie++ parsing part?

omnium21:
> @omnium21 I think it's good to be merged. Should I wait for the bonnie++ parsing part?

The bonnie parsing from the old test no longer works, and it doesn't look like a trivial fix. I'll push a quick update to tee the results rather than redirect them to the log file, so they appear in the terminal output and think about the parsing another time...

omnium21:
I've pushed an updated version, including a check to make sure the device isn't mounted before we attempt to format or mount it in the test.
If mounted, format fails.
If mounted, mount skips.
I'm running it in LAVA now...
https://validation.linaro.org/scheduler/job/1980712

omnium21:
@mwasilew , I think it's ready now.

PR #171, opened 2020-06-10
test-runner: change pexpect.spawn to spawnu

mwasilew:
After migration to python3 pexpect started failing on remote tests. This
can be fixed by moving to 'spawnu' so all input is considered utf-8


vishalbhoj:
looks good to me !

PR #180, opened 2020-07-03
Some improvement on the tradefed  test

liuyq:
1. fixed the problem that not all xml un-support characters removed.
2. continue to try to upload the result/log files to archive site even the tradefed test itself failed.

Test job which get the result successfully:
https://validation.linaro.org/scheduler/job/2048126

The old jobs:
https://validation.linaro.org/scheduler/job/2045998 failed to get the result successfully
https://validation.linaro.org/scheduler/job/2045733 not have the attachment uploaded

### DIFF
--- a/automated/android/tradefed/result_parser.py
+++ b/automated/android/tradefed/result_parser.py
@@ -57,7 +57,12 @@ class TradefedResultParser:
             ):
                 etree_content = etree_content[:mstart] + etree_content[mend:]
                 endpos = len(etree_content)
-            pos = mend
+                # next time search again from the same position as this time
+                # as the detected pattern was removed here
+                pos = mstart
+            else:
+                # continue from the end of this match
+                pos = mend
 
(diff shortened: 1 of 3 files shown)


automated/android/tradefed/result_parser.py
```diff
@@ -57,6 +57,11 @@ def parse(self, xml_file):
             ):
                 etree_content = etree_content[:mstart] + etree_content[mend:]
                 endpos = len(etree_content)
+            # next time search again from the same position as this time
+            # as the detected pattern was removed here
+            pos = mstart
+        else:
```
mwasilew, line 63: is this properly indented? It looks like 'else' has the same indentation as 'while'. Seems weird to me as pos has no meaning outside of while loop

liuyq: sorry, it's not correct. updated already.

PR #185, opened 2020-07-14
kselftest: check for run script

khilman:
The kselftest.sh wrapper currently assumes that the kselftest tarball
was created with a top-level "kselftest" directory, which builds
in an assumption about how the tarball was created.

To fix, test to see if the top-level run-script is present in the
current dir before trying to "cd kselftest"

Signed-off-by: Kevin Hilman <khilman@baylibre.com>


mwasilew:
@nareshkamboju could you check this? It looks OK for me. Doesn't change the current behavior.

PR #187, opened 2020-07-23
armnn-unit-tests: creating test definitions for armnn unit tests

K1504296:
automated: linux: add the armnn unit tests

Adding armnn-unit-tests files to automated/linux/armnn/ in order to facilitate deployment to 64bit and 32bit devices in the LAVA lab as part of the armnn ci loop. This will deploy Deploy ArmNN as well as it's build UnitTests to a board which does not already contain them. If a board already has ArmNN installed it will also allow for the UnitTests to be automated and run. 

Signed-off-by: Theodore Grey <theodore.grey@linaro.org>

### DIFF
--- a/automated/linux/armnn/armnn-unit-tests.sh
+++ b/automated/linux/armnn/armnn-unit-tests.sh
@@ -1,0 +1,41 @@
+#!/bin/sh
+
+. ../../lib/sh-test-lib
+OUTPUT="$(pwd)/output"
+
+usage() {
+    echo "Usage: $0 [-s <true|false>]" 1>&2
+    exit 1
+}
+
+while getopts "s:a:d:" o; do
+  case "$o" in
+    s) SKIP_INSTALL="${OPTARG}" ;;
+    a) ARMNN_TARBALL="${OPTARG}" ;;
+    d) TEST_DIR="${OPTARG}" ;;
+    *) usage ;;
+  esac
+done
+
+! check_root && error_msg "You need to be root to run this script."
+create_out_dir "${OUTPUT}"
+
+pkgs="ntp wget gcc g++ python3 python3-pip"
+dhclient
+install_deps "${pkgs}" "${SKIP_INSTALL}"
+
+if [[-n ${ARMNN_TARBALL} ]]; then
+    mkdir -p ${TEST_DIR}
+    cd ${TEST_DIR}
+    wget -O armnn.tar.xz "${ARMNN_TARBALL}"
+    tar xf armnn.tar.xz
+    export BASEDIR="/home/buildslave/workspace/armnn-ci-build"
+fi
+
+if [[ ${SKIP_INSTALL} == false ]]; then
+    cd $BASEDIR/build
+    ln -s libprotobuf.so.15.0.0 ./libprotobuf.so.15
+    export LD_LIBRARY_PATH=`pwd`
+    chmod a+x UnitTests
+fi
+lava-test-case ArmNN-Unit-Tests --shell ./UnitTests
--- a/automated/linux/armnn/armnn-unit-tests.yaml
+++ b/automated/linux/armnn/armnn-unit-tests.yaml
@@ -1,0 +1,25 @@
+metadata:
+    format: Lava-Test Test Definition 1.0
+    name: armnn-unit-tests
+    description: "Carry out armnn unit tests."
+    maintainer:
+        - theodore.grey@linaro.org
+    os:
+        - debian
+    scope:
+        - functional
+    devices:
+        - synquacer
+        - dragonboard-845c
+        - hi960-hikey
+        - stm32mp1
+params:
+    # if you need to run for 32 bit devices use:
+    # 'https://snapshots.linaro.org/componenets/armnn-32bit/latest/armnn-32.tar.xz'
+    ARMNN_TARBALL: 'https://snapshots.linaro.org/componenets/armnn/latest/armnn.tar.xz'
+    TEST_DIR: '/usr/local/bin'
+    SKIP_INSTALL: false
+run:
+    steps:
+        - cd ./automated/linux/armnn/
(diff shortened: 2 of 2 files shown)


automated/linux/armnn/armnn-unit-tests.sh
```diff
@@ -0,0 +1,43 @@
...
+
+if [ "$abi" == "armeabi" ]
+then
+    mkdir /usr/local/bin && cd /usr/local/bin && \
+    wget "${LINK_SNAPSHOT_32BIT}" && tar xf armnn-32bit.tar.xz && \
+    export BASEDIR="/home/buildslave/workspace/armnn32-ci-build"
+else
+    wget "${LINK_SNAPSHOT_64BIT}" && tar xf armnn.tar.xz && \
+    export BASEDIR="home/buildslave/workspace/armnn-ci-build"
+
```
roxell, line 38: missing an "fi" ?

automated/linux/armnn/armnn-unit-tests.yaml
```diff
@@ -0,0 +1,23 @@
...
+    scope:
+        - functional
+    devices:
+        - synquacer
+        - dragonboard-845c
+        - hi960-hikey
+        - stm32mp1
+params:
+    LINK_SNAPSHOT_64BIT: 'https://snapshots.linaro.org/componenets/armnn/latest/armnn.tar.xz'
+    LINK_SNAPSHOT_32BIT: 'https://snapshots.linaro.org/componenets/armnn-32bit/latest/armnn-32.tar.xz'
```
roxell, line 18: maybe just have one variable like ARMNN_TARBALL and point that to either armnn-32.tar.xz or armnn.tar.xz... then add the other link as a comment.

automated/linux/armnn/armnn-unit-tests.yaml
```diff
@@ -0,0 +1,23 @@
...
+    SKIP_INSTALL: false
+run:
+    steps:
+        - cd ./automated/linux/armnn/
+        - ./armnn-unit-tests.sh -s "${SKIP_INSTALL}" -a "${LINK_SNAPSHOT_64BIT}" -b "${LINK_SNAPSHOT_32BIT}"
```
roxell, line 23: Then we can only have one of the flags...
What happens if I use a rootfs that already have my armnn preinstalled? then I want to  be able to run the test too... but not build and stuff...

automated/linux/armnn/armnn-unit-tests.sh
```diff
@@ -0,0 +1,43 @@
...
+create_out_dir "${OUTPUT}"
+
+pkgs="ntp wget gcc g++ python3 python3-pip"
+dhclient
+install_deps "${pkgs}" "${SKIP_INSTALL}"
+
+#(GET WHICH PLATFORM)
+detect_abi
+
+if [ "$abi" == "armeabi" ]
```
roxell, line 30: I think you can drop the if statement if you pass '-O my-armnn.tar.xz' to wget...

Then you can do 'tar xf my-armnn.tar.xz'
BASEDIR can't that be the same also?

automated/linux/armnn/armnn-unit-tests.sh
```diff
@@ -0,0 +1,43 @@
...
+    wget "${LINK_SNAPSHOT_32BIT}" && tar xf armnn-32bit.tar.xz && \
+    export BASEDIR="/home/buildslave/workspace/armnn32-ci-build"
+else
+    wget "${LINK_SNAPSHOT_64BIT}" && tar xf armnn.tar.xz && \
+    export BASEDIR="home/buildslave/workspace/armnn-ci-build"
+
+cd $BASEDIR/build
+ln -s libprotobuf.so.15.0.0 ./libprotobuf.so.15
+export LD_LIBRARY_PATH=`pwd`
+chmod a+x UnitTests
```
roxell, line 42: these four lines shouldn't be done if we set 'SKIP_INSTALL=true', and don't pass in any tarball right (which implies that the armnn is already in the rootfs...

automated/linux/armnn/armnn-unit-tests.sh
```diff
@@ -0,0 +1,43 @@
...
+./UnitTests
```
roxell, line 43: how does the output look from this ?
is it in a format that LAVA can parse ?

K1504296: I've added "lava-test-case ArmNN-Unit-Tests --shell ./UnitTests" to resolve this.

automated/linux/armnn/armnn-unit-tests.sh
```diff
@@ -24,20 +24,18 @@ pkgs="ntp wget gcc g++ python3 python3-pip"
...
-else
-    wget "${LINK_SNAPSHOT_64BIT}" && tar xf armnn.tar.xz && \
-    export BASEDIR="home/buildslave/workspace/armnn-ci-build"
-
-cd $BASEDIR/build
-ln -s libprotobuf.so.15.0.0 ./libprotobuf.so.15
-export LD_LIBRARY_PATH=`pwd`
-chmod a+x UnitTests
-./UnitTests
+if [[-z ${ARMNN_TARBALL} ]]; then
```
roxell, line 27: I'm so sorry, this is my fault... I think it should be '-n' and not '-z'.
       -n STRING
              the length of STRING is nonzero
       -z STRING
              the length of STRING is zero

roxell:
Can you please re-word the changelog.
maybe something like this:
"
automated: linux: add the armnn unit tests

<add a bit of description what it is and do>

Signed-off-by: Theodore Gray <theodore.grey@linaro.org>
"

automated/linux/armnn/armnn-unit-tests.sh
K1504296: Ah yeah you're right, good catch

roxell:
Awesome work!
LGTM.

PR #188, opened 2020-07-24
boottime:  add support aosp master version

liuyq:
collect ro.boottime properties and replace dc with awk tool
and changed to get information from logcat dmesg buffer
instead of dmesg log which might not have all logs

Tested with following jobs:
db845c: https://lkft.validation.linaro.org/scheduler/job/1598559
hikey960: https://validation.linaro.org/scheduler/job/2073798
x15: https://lkft.validation.linaro.org/scheduler/job/1596656

Problems:
1. launcher displayed log not printed for the second boot
2. it's not stable to reboot on lava, often failed to get
   the result for the second boot because of reboot problem.

   need to check from both lava and device side

### DIFF
--- a/automated/android/boottime/boottime.sh
+++ b/automated/android/boottime/boottime.sh
@@ -16,13 +16,14 @@ usage() {
     exit 1
 }
 
-while getopts ":S:s:t:o:n:" o; do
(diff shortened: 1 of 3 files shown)


mwasilew:
Please rebase. The problem with armnn is fixed on master branch

PR #189, opened 2020-07-24
apk-automation: update to work with aosp master branch

liuyq:
Tested locally with hikey960 aosp-master-throttled-copped#6640860 build


liuyq:
The check failures seem not caused by this change?

PR #197, opened 2020-08-04
igt: Check Chamelium IP address before we go further

arthur-she:
The Chamelium IP address comes from device dict. It has to be ready
before we do the further tests.


nareshkamboju:
@arthur-she Please post tested LAVA job with this change.

arthur-she:
Here is the test job.
https://lkft.validation.linaro.org/scheduler/job/1644146#L4134
There are many "SKIP" in the test results, but that is irrelevant.

PR #200, opened 2020-08-31
fix lemp test on debian

vishalbhoj:
Signed-off-by: Vishal Bhoj <vishal.bhoj@linaro.org>


vishalbhoj:
Tested here: https://validation.linaro.org/results/2117609

PR #201, opened 2020-08-31
Preempt rt

igaw:
This PR contains updates to the rt-tests testsuite. I am running this version for a while locally and feel comfortable to share it with upstream. See also https://github.com/kernelci/kernelci-core/issues/474

Changes:

- Add background command feature to all rt-tests. It utilizes a very simple form of background job control. I tried  to keep it a simple as possible and not trying to over engineer it at this point. It works reliable so far and I  think it's a good starting point.
- Silence the very verbose output of pi-stress and signaltest. All other rt-tests don't have a quiet option yet.  I'm planing to update and streamline all rt-tests in upstream soon. It's very annoying the small difference in each program.
~~- Remove the MAX_LATENCY argument again. It was not a good idea to mix executation/controlling of the test jobs and the performance measurement. This is done better in post processing. Rip it out again. I kept the parse_rt_tests_results.py for all rt-tests. The output format every test is slightly different and I was not able to get it working in a pure shell version. Until we have a better output format from upstream (yep planing to add this), let's stick with the current script.~~

### DIFF
--- a/automated/linux/signaltest/signaltest.sh
+++ b/automated/linux/signaltest/signaltest.sh
@@ -12,18 +12,20 @@ PRIORITY="98"
 THREADS="2"
 MAX_LATENCY="100"
 DURATION="1m"
+BACKGROUND_CMD=""
 
 usage() {
-    echo "Usage: $0 [-r runtime] [-p priority] [-t threads] [-m latency]" 1>&2
+    echo "Usage: $0 [-r runtime] [-p priority] [-t threads] [-m latency] [-w background_cmd]" 1>&2
     exit 1
 }
 
-while getopts ":p:t:D:m:" opt; do
+while getopts ":p:t:D:m:w:" opt; do
     case "${opt}" in
         p) PRIORITY="${OPTARG}" ;;
         t) THREADS="${OPTARG}" ;;
 	D) DURATION="${OPTARG}" ;;
 	m) MAX_LATENCY="${OPTARG}" ;;
+	w) BACKGROUND_CMD="${OPTARG}" ;;
         *) usage ;;
     esac
 done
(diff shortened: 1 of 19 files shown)


automated/linux/signaltest/signaltest.sh
```diff
@@ -42,7 +42,7 @@ fi
 
 background_process_start bgcmd --cmd "${BACKGROUND_CMD}"
 
-"${binary}" -D "${DURATION}" -m -p "${PRIORITY}" -t "${THREADS}" \
+"${binary}" -q -D "${DURATION}" -m -p "${PRIORITY}" -t "${THREADS}" \
```
igaw, line 43: I think we should drop this patch at this point. Currently, there is a problem with the reporting when used the quiet argument:

> When the quiet option is used don't print all threads stats. The first
> thread will pause every 16 cycles for 10000us. This will show up in
> the max values for all other threads when the final results are
> printed:
> 
>  # signaltest -q  -D 2s
>  T: 0 ( 4517) P: 0 C:   3121 Min:      5 Act:    8 Avg:    8 Max:      19
>  T: 1 ( 4518) P: 0 C:   3121 Min:      5 Act:10065 Avg:  639 Max:   10073
> 
> We could also remove the sleep but then the system gets fully loaded
> by the test. Furthermore, we would keep the path pretty hot and that's
> not ideal if one wants to test the eratic signal behavior. So only
> consider the first thread for the stats.


https://lore.kernel.org/linux-rt-users/20200901154657.30198-5-dwagner@suse.de/

mwasilew:
@igaw this looks OK to me with the exception of MAX_LATENCY commit. Maybe it should go to a separate PR?

@roxell, @khilman could you take a look and add any objections if you have?

khilman (approved):
LGTM, tested in KernelCI.

igaw:
> @igaw this looks OK to me with the exception of MAX_LATENCY commit. Maybe it should go to a separate PR?

Sure, no problem. I'll prepare a new PR for the feature removal.

mwasilew:
Merged. Thank you for contribution!

PR #203, opened 2020-09-04
rt-tests: Remove MAX_LATENCY argument

igaw:
Remove the MAX_LATENCY argument again. It was not a good idea to mix execution/controlling of the test jobs and the performance measurement. This is done better in post processing. Rip it out again. I kept the parse_rt_tests_results.py for all rt-tests. The output format every test is slightly different and I was not able to get it working in a pure shell version. Until we have a better output format from upstream (yep planing to add this), let's stick with the current script.

### DIFF
--- a/automated/linux/ptsematest/ptsematest.sh
+++ b/automated/linux/ptsematest/ptsematest.sh
@@ -8,18 +8,16 @@ OUTPUT="${TEST_DIR}/output"
 LOGFILE="${OUTPUT}/ptsematest.log"
 RESULT_FILE="${OUTPUT}/result.txt"
 DURATION="5m"
-MAX_LATENCY="100"
 BACKGROUND_CMD=""
 
 usage() {
-    echo "Usage: $0 [-D duration] [-m latency] [-w background_cmd]" 1>&2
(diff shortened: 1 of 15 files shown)


automated/linux/ptsematest/ptsematest.sh
```diff
@@ -8,18 +8,16 @@ OUTPUT="${TEST_DIR}/output"
...
-MAX_LATENCY="100"
 BACKGROUND_CMD=""
 
 usage() {
-    echo "Usage: $0 [-D duration] [-m latency] [-w background_cmd]" 1>&2
+    echo "Usage: $0 [-D duration] [-w background_cmd]" 1>&2
     exit 1
 }
 
 while getopts ":D:m:w:" opt; do
```
igaw, line 18: Forgot to remove the 'm:' here. Let me update it.

mwasilew:
@roxell could you please take a look? I'm OK with the change. It makes sense to always pass the test with measurement as there is no a priori knowledge about the historical trend or expected result. So I'm OK removing MAX_LATENCY as it would need to be ignored in postprocessing anyway.

roxell:
LGTM

PR #205, opened 2020-09-09
automated/linux: glmark2 allow to specify binary

alimon:
Newer versions of glmark2 has binaries like glmark2-es2
and glmark2-es2-drm.

Signed-off-by: Aníbal Limón <anibal.limon@linaro.org>


mwasilew:
It seems there is a hidden assumption that glmark2 is already installed. Could you also add install.sh script to the test?

alimon:
@mwasilew Right, I installed previously glmark2 in OE RPB for Debian there is no package in main repositories.

PR #206, opened 2020-09-11
igt: Check target network status before go further

arthur-she:
Fix the test result will be misleading when the network is unavailable.
1. Stop testing if the network is not available
2. Power cycle Chamelium before running igt Chamelium test


arthur-she:
The problematic test job is [this](https://lkft.validation.linaro.org/scheduler/job/1750923#L2241)
The network is unavailable, however, the test result of "Ping-Chamelium" was a pass.

Test job for testing:
1. [network is available](https://lkft.validation.linaro.org/scheduler/job/1755446)
2. [network is unavailable](https://lkft.validation.linaro.org/scheduler/job/1755444#L1735)

PR #208, opened 2020-09-15
kselftest: Improve logging output and reporting accuracy

kees:
This updates the kselftest wrapper to more accurately report the test results. Changes include:

- skipping tests based on regular expression so that the full test path is used for matches
- parsing test output for kTAP
- including externally skipped tests in results
- avoiding log mangling when post-processing gives saner results
- actually send result.txt to LAVA


mwasilew:
@nareshkamboju Is it OK for you to add perl as dependency to kselftest?

roxell:
LGTM, apart from the mixing tab and spaces.

nareshkamboju:
@mwasilew testing this patch set on our OE builds 
https://lkft.validation.linaro.org/scheduler/job/1766015

khilman:
This is great, thanks @kees !

While I'd rather see the parsing in python than perl, beggars cannot be choosers. :)  KernelCI debian rootfs baseline images already include perl, so we should survive.

kees:
Is rather do this in portion too, but yes, I chose Perl because it was already present in most cases. Further down the road, this should probably have a full TAP parser in LAVA itself, but since kselftest is really the only user currently, I figure that can be for the future.

nareshkamboju:
@mwasilew @kees  I have tested your results parser and results looks good to me.

test job link,
https://lkft.validation.linaro.org/results/1766015

automated/linux/kselftest/kselftest.yaml
```diff
@@ -48,10 +48,4 @@ run:
...
         - cd ./automated/linux/kselftest/
         - ./kselftest.sh -t "${TESTPROG}" -s "${SKIP_INSTALL}" -u "${TESTPROG_URL}" -L "${SKIPLIST}" -S "${SKIPFILE}" -b "${BOARD}" -g "${BRANCH}" -e "${ENVIRONMENT}" -p "${KSELFTEST_PATH}"
-
-parse:
-    pattern: "^selftests:\\s(?P<test_case_id>\\S+)\\s[[](?P<result>(FAIL|PASS|SKIP))[]]"
-    fixupdict:
-        FAIL: fail
-        PASS: pass
-        SKIP: skip
+        - ../../utils/send-to-lava.sh ./output/result.txt
```
nareshkamboju, line 51: Kselftest runs tests in sequence and when we see kernel crash at middle of the test run we loose all the results with this new parser at the end.

The older parser was run time parser for each test case.

kees: IIUC I could change this to perform the "send-to-lava" logic for each test?

nareshkamboju: yes. 
Please refer this commit,
https://github.com/Linaro/test-definitions/commit/0fd41f9b0b65c17782e1a4b8cb5826b3af6eaec6#diff-e2e0fe7e0368d547b14606c0b5e209cb

kees: Ah, yes, okay. I understand now. I think I was looking at the rest runner and thinking it was struggle local, but if this is getting collected over the console or whatever, yes, that makes sense. I'll need to think about this a bit more, since correcting the kselftest test names is not trivial.

One question, though: it seems in many cases that the kernel's dmesg output is interleaved with test execution output which makes reading log codifying at best and at worst makes the test output parsing non-deterministic. It seems that LAVA is seeing the same interleaving, yes? I think that'll need fixing as well (i.e. something needs to be doing proper line-at-a-time interleaving of test output and dmesg output).

Thanks for the review! I'll get this series updated...

automated/linux/kselftest/kselftest.sh
```diff
@@ -110,15 +110,42 @@ fi
...
+        }
+        # Any appearance of the SKIP is a skip.
+        if ($skip eq "SKIP") {
+            $result="skip";
+        } elsif ($not eq "not ") {
+            $result="fail";
+        } else {
+            $result="pass";
+        }
+	print "$test $result\n";
```
nareshkamboju, line 138: with your patch applied i see great improvement in test count increased from 375 to 934 in total.

Before patch applied:
PASS    FAIL   TOTAL
173      168     375



After patch applied:
PASS  FAIL   TOTAL
702     186    934
https://lkft.validation.linaro.org/results/1766015

mwasilew:
@nareshkamboju I'd say it's your call. I'm OK merging the patch as it is. The problems we might run into is loosing all results in the event of kernel crash. On the positive side the change allows to collect all (or almost all) results. Current version can't do that. If you're OK with that let's go ahead. 

@kees the problem of test output mixed with dmesg is a real one. There is no good solution I know of. You could turn dmesg off before starting the test but it might mask some errors. I guess the only proper solution is running the test in different terminal then dmesg is sent to. But this complicates setup a lot. So for now I'm afraid we have to live with the issue. 

As for LAVA support for parsing kTAP/TAP13 or similar please submit patches to LAVA(git.lavasoftware.com). I'm happy to review them. Given the amount of code in this patch required to parse the output this shouldn't be too hard.

nareshkamboju:
@mwasilew  and @kees  we will work on improvements in next set of patches.

PR #209, opened 2020-09-15
Glmark2 fixes and new deqp-runner test

alimon:


automated/linux/deqp-runner/deqp-runner.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+run:
+    steps:
+        - export DISPLAY=${DISPLAY}
+        - export EGL_PLATFORM=${EGL_PLATFORM}
+        - export RESULT_FILE=result.txt
+        - cd ./automated/linux/deqp-runner
+        - set +e
+        - deqp-runner --deqp ${DEQP_BIN} --output $RESULT_FILE ${DEQP_CASES} ${DEQP_FAIL} ${DEQP_EXCLUDE} ${DEQP_RUNNER_OPTIONS} ${DEQP_RUNNER_JOBS} -- ${DEQP_OPTIONS}
+        - DEQP_EXITCODE=$?
```
mwasilew, line 37: the whole block below can be replaced with ../../utils/send-to-lava.sh ${RESULT_FILE}. It would also be better to convert the code above this comment to a shell script so it can be executed without yaml file (meaning without LAVA or test-runner.py).

mwasilew (changes requested):
deqp-runner-lava.py duplicates features already present in send-to-lava.sh.

liuyq:
@alimon Interested in the test results on dragonboard845c board, are there any initial results already there that could could be shared?

alimon:
> @alimon Interested in the test results on dragonboard845c board, are there any initial results already there that could could be shared?

@liuyq Here are the results for 845c using 5.7.19 qcomlt rel branch, https://qa-reports.linaro.org/qcomlt/linux-release/build/480/?failures_only=false#!#test-results

liuyq:
> > @alimon Interested in the test results on dragonboard845c board, are there any initial results already there that could could be shared?
> 
> @liuyq Here are the results for 845c using 5.7.19 qcomlt rel branch, https://qa-reports.linaro.org/qcomlt/linux-release/build/480/?failures_only=false#!#test-results

@alimon Thanks, we have huge failures reported by the android cts deqp test, will compare between them.

automated/linux/deqp-runner/deqp-runner.yaml
alimon: @mwasilew I can create a shell script to execute the test, regarding the send-to-lava.sh change the results python script has different regex [1] and has a mapping inside for results [2].

[1] https://github.com/Linaro/test-definitions/pull/209/commits/290d4afdf2bfe9730356cbc35c8755063140f7db#diff-49ed82b07280d7f16b68a1c465b0a882R40
[2] https://github.com/Linaro/test-definitions/pull/209/commits/290d4afdf2bfe9730356cbc35c8755063140f7db#diff-49ed82b07280d7f16b68a1c465b0a882R28

mwasilew: @alimon I would prefer to have the RESULT_FILE in a standardized format. If you need to translate results, please save the intermediate file with a different name. Please check iozone.sh as example.

automated/linux/deqp-runner/deqp-runner.yaml
```diff
@@ -0,0 +1,48 @@
...
+        - set -e
+        - test_set_name="$(basename ${DEQP_BIN})"
+        - lava-test-set start "$test_set_name"
```
mwasilew, line 40: please use sh-test-lib functions instead of calling lava-* helpers directly. This test should also run nicely standalone and with test-runner.py.

alimon: @mwasilew There is no lava-test-set function inside sh-test-lib.

mwasilew: so add it :)

automated/linux/deqp-runner/deqp-runner.yaml
alimon: @mwasilew I made the changes requested to use a result file with standardized format, I noticed that increase the running time 1 hour, 7 minutes [1] vs 1 hour, 43 minutes [2]. 

Sometime ago I found some high cpu consumption parsing results using pexpect in lava-dispatcher may be is the currently issue. For now I'm fine with the current times.

[1] https://validation.linaro.org/scheduler/job/2163739
[2] https://validation.linaro.org/scheduler/job/2165199

mwasilew:
Thank you. Did you try it with test-runner.py? I guess it might be unfamiliar with 'test-set' reporting. Anyway, it's not the main use case. I'll check if it works later. If not, we can fix it.

PR #214, opened 2020-09-25
crypto: basic tests

montjoie:
Implement a simple crypto testsuite which try to loads all crypto
algorithm and so force selftests to run.

This test suite will check for errors and then a pass/fail will be assigned for each crypto algorithm.

### DIFF
--- a/automated/linux/crypto/crypto.sh
+++ b/automated/linux/crypto/crypto.sh
@@ -1,0 +1,105 @@
+#!/bin/sh
+
+. ../../lib/sh-test-lib
+
+start_test() {
+	echo "=============================================="
+	echo "$1"
+	echo "=============================================="
+	dmesg_capture_start
+}
+
+result() {
+	result=$1
+	shift
+
+	dmesg_capture_result
+	if [ "$result" = 'FAIL' ];then
+		report_fail "$1"
+	fi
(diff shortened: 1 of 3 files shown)


mwasilew:
This looks good. Please fix the shellcheck problems:
https://travis-ci.com/github/Linaro/test-definitions/jobs/390810053#L1211

automated/linux/crypto/crypto.sh
```diff
@@ -0,0 +1,105 @@
...
+# tcrypt generate error -2 for non-present algs
+start_test "Verify crypto errors"
+dmesg | grep -vE 'is unavailable$|[[:space:]]-2$|This is intended for developer use only|alg: No test for stdrng' |grep alg:
+RET=$?
+if [ $RET -eq 0 ];then
+	result 1 "crypto-error-log"
+else
+	result 0 "crypto-error-log"
+fi
+
```
mwasilew, line 79: would it make sense to do 'stop_test' here and turn dmesg print to console back? This will leave the system in good shape for the next test.

montjoie: Yes, I forgot that all my test suite use the dmesg trick so it was not a problem.
But for yours it is.
I will fix it

PR #222, opened 2020-10-03
test-runner: make sure the measurement is a number

ndechesne:
In one of our tests, we had:
openssl-speed,openssl-version,pass,1.1.1g,version,SKIP_INSTALL=False

And when submitting to Squad, it failed because the measurement was
not a 'number' (e.g. 1.1.1g). With this patch, we make sure that all
measurements reported to Squad are integers or float.

Signed-off-by: Nicolas Dechesne <nicolas.dechesne@linaro.org>

### DIFF
--- a/automated/utils/test-runner.py
+++ b/automated/utils/test-runner.py
@@ -758,7 +758,10 @@ class ResultParser(object):
                         if result_match:
                             data['result'] = result_match.group(1)
                         if measurement_match:
-                            data['measurement'] = measurement_match.group(1)
+                            try:
+                                data['measurement'] = float(measurement_match.group(1))
+                            except ValueError as e:
+                                pass
                         if units_match:
                             data['units'] = units_match.group(1)
 


automated/utils/test-runner.py
```diff
@@ -804,7 +804,7 @@ def send_to_qa_reports(self):
         tests = {}
         metrics = {}
         for metric in self.metrics:
-            if metric['measurement'] != "":
+            if metric['measurement'].replace('.','',1).isdigit():
```
mwasilew, line 807: hmm, I'm not sure this will work as intended:
```
>>> a = "4"
>>> a.isdigit()
True
>>> a = " 4"
>>> a.isdigit()
False
>>> a = "1.1"
>>> a.isdigit()
False
>>> a = ""
>>> a.isdigit()
False
```
Maybe a better way would be trying to convert string to float and catch exception if it fails?

ndechesne: you should test with .replace('.','',1).isdigit() so that it would be True for float.

mwasilew: hmm, in the commit message you wrote you had value as 1.1.1g. So when I replace 1 "." with "" I will end up with 1.1
```
>>> a = "1.1.1"
>>> a.replace(".", "", 1).isdigit()
False
```

ndechesne: 1.1.1g is not a float (because of the 'g')
1.1.1 is not a float neither
1.1 should be treated as a float. 

so this patch would be to catch things like "1" or "1.1" and nothing else

mwasilew: ah, now I got what you meant. I still think converting to float is slightly better idea:
```
>>> a = "1.1"
>>> b = float(a)
>>> b
1.1
>>> a = "1.1.1"
>>> b = float(a)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: could not convert string to float: '1.1.1'
```

So maybe sth like this:
```
if metric['measurement'] != "":
    try: 
        metrics["{}/{}".format(self.test['test_name'], metric['test_case_id'])] = float(metric['measurement'])
    except ValueError as e:
        pass
```

ndechesne: i agree it's better. but we probably want to record metric['result'] if the measurement is not valid.

mwasilew: Yes, I missed that. The except block should contain:
```
tests["{}/{}".format(self.test['test_name'], metric['test_case_id'])] = metric['result']
```

ndechesne: @mwasilew , I decided to to it differently, and check for float when we parse the the output. i think it's even cleaner. I tested with 1.1.1h, 1.2, and 1.1.2, and it works as expected.

mwasilew: Should we abandon this? IMHO it's still important to add.

ndechesne: I pushed an updated patch. the implementation is different, but it fixes the problem. check the new patch.

PR #223, opened 2020-10-03
kunit: Adding kunit tests

nareshkamboju:
KUnit is a lightweight unit testing and mocking framework for the Linux kernel.

KUnit consists of a kernel component, which provides a set of macros for easily
writing unit tests. Tests written against KUnit will run on kernel boot if
built-in, or when loaded if built as a module.

### DIFF
--- a/automated/linux/kunit/kunit.sh
+++ b/automated/linux/kunit/kunit.sh
@@ -1,0 +1,87 @@
+#!/bin/sh
+
+# shellcheck disable=SC1091
+. ../../lib/sh-test-lib
+OUTPUT="$(pwd)/output"
+RESULT_FILE="${OUTPUT}/result.txt"
+export RESULT_FILE
+TEST_LOG="${OUTPUT}/test_log.txt"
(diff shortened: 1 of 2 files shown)


nareshkamboju:
Tested on LAVA and getting 90 test cases results.
https://lkft.validation.linaro.org/results/1814997/1_kunit-tests?search=&length=100#table

roxell:
LGTM, just a few questions.
1. what will happen if a test adds "# SKIP" to the end of the test result line? 
'not ok test_1 # SKIP'
2. What happens if I build the tests as a module don't we have to run them first and them first in the .sh script and then parse?

nareshkamboju:
@roxell  Thanks for the review,
Since you have provided the SKIP scenario example. I will address that.
Building it as module is also one of the use case, I will add that too.

- Naresh

roxell:
@nareshkamboju I'm being fine if you add that in a follow up commit after this is merged.

nareshkamboju:
@roxell the skip functionality added now.
The pending usecase is modprobe kunit.ko

nareshkamboju:
This patch is ready to merge.

PR #231, opened 2020-10-15
kselftest: Adjust skip logic to handle latest kselftest logic

kees:
Instead of manipulating the run script, censor the list of tests.

Signed-off-by: Kees Cook <keescook@chromium.org>

### DIFF
--- a/automated/linux/kselftest/kselftest.sh
+++ b/automated/linux/kselftest/kselftest.sh
@@ -193,28 +193,13 @@ if [ -f "${SKIPFILE}" ] &&  [ -z "${SKIPLIST}" ]; then
     done < "${SKIPFILE}"
 fi
 
-echo "========================================"
+cp kselftest-list.txt kselftest-list.txt.orig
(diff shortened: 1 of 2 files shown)


nareshkamboju:
The upstream change is now in Linux -next tree master branch. This patch works only on that tree and branch.
So we have to merge this patch into test-definitions next branch and use it for Linux next master branch.

For mainline and stable-rc we have to use kselftest test definitions from master branch.

mwasilew:
There is no test-definitions -next branch. It has to wait till the patch reaches mainline and eventually stable trees.

kees:
> There is no test-definitions -next branch. It has to wait till the patch reaches mainline and eventually stable trees.

Another idea: it could handle both. Let me cook that up...

aliceinwire:
KernelCI kselftest skip logic changed, 
this pull request is handled differently and can now be closed.

kees:
Abandoned

PR #232, opened 2020-10-18
Igt tuning

arthur-she:
1. Remove command which is used to delete Chamelium logs
2. Force python to use utf-8 for stdout/stderr encoding


arthur-she:
The problematic test job for Python encoding
https://lkft.validation.linaro.org/scheduler/job/1854882#L8867

The test job which adapts this patch
https://lkft.validation.linaro.org/scheduler/job/1859941

PR #233, opened 2020-10-19
Fix fwts

roxell:


automated/linux/fwts/fwts.sh
```diff
@@ -182,5 +182,9 @@ if [ "${SKIP_INSTALL}" = "true" ] || [ "${SKIP_INSTALL}" = "True" ]; then
 else
 	install
 fi
-get_tests
+
+which fwts
```
mwasilew, line 191: hmm, I wonder why shellcheck didn't complain. You could rewrite this as
```
if ! (which fwts); then get_tests; fi
```

automated/linux/fwts/fwts.sh
```diff
@@ -168,7 +170,8 @@ parse_fwts_test_results() {
 
 run_test() {
 
-	fwts "${TESTS}" - 2>&1 | tee -a "${RESULT_LOG}"
+        # shellcheck disable=SC2086
```
mwasilew, line 175: I understand why this is needed. Could you add a comment explaining it?

mwasilew (changes requested):

roxell:
oops closing this one its in the new #236

PR #234, opened 2020-10-20
Fix git clone

roxell:


mwasilew:
@roxell why create new test-definition for cloning LTP from git while it copies most of the original ltp code? Can't the support for git cloning be added to the original one?

PR #235, opened 2020-10-21
Rteval fixes

roxell:


automated/lib/sh-test-lib
```diff
@@ -662,3 +662,40 @@ dmesg_capture_result() {
 	dmesg --console-on
 	echo "=============================================="
 }
+
+# shellcheck disable=SC2039
+get_test_program() {
```
mwasilew, line 667: Is this going to be used in other tests?

mwasilew (changes requested):
I'd like some more verbose commit message in the merge commit. The changes seem to make sense, but some justification would be nice to see.

mwasilew (approved):

PR #238, opened 2020-10-21
Run ltp from git

roxell:


mwasilew:
Why do we need a new YAML file? It seems to me that the previous YAML is good enough.

mwasilew:
This changes the logic a little bit but I'm OK with the change.

PR #242, opened 2020-11-01
boottime.yaml: remove the passing of ANDROID_SERIAL

liuyq:
to workaround the naming conflict with the docker method.
with the docker method, ANDROID_SERIAL defined in the yaml file
or in the boottime.sh will cause the adb get-serialno failed to
get the device serial number.

If there is a need to specify the serial number for the script here,
plese export ANDROID_SERIAL explicitly before running the scripts.

### DIFF
--- a/automated/android/boottime/boottime.sh
+++ b/automated/android/boottime/boottime.sh
@@ -1,7 +1,6 @@
 #!/bin/sh -e
 # shellcheck disable=SC1091
 
-ANDROID_SERIAL=""
 BOOT_TIMEOUT="300"
 OPERATION="COLLECT"
 COLLECT_NO="1"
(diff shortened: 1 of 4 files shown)


mwasilew:
@liuyq could you add commit message to the YAML file metadata/description field? This way it will get rendered in the documentation.

liuyq:
> @liuyq could you add commit message to the YAML file metadata/description field? This way it will get rendered in the documentation.

@mwasilew where can I see the rendered documentation? it would be better for me to check the document first, and update the description with that as reference.

mwasilew:
https://test-definitions.readthedocs.io/en/latest/automated/android/boottime/boottime.html

liuyq:
@mwasilew just updated the description, please help to check again.

PR #243, opened 2020-11-04
android: fix potential issue in tradefed debug message

mwasilew:


liuyq:
@mwasilew  could you please help revert the commit 48f0664c1c301928d46762c7a47ec036dd064116 first, and do this change on that? it seems we still need the error_fatal to be run in the yaml file, otherwise the adb disconnect problem won't be reported.

mwasilew:
Sure, I'll add it to this patch

liuyq:
thanks, looks good to me. +1

PR #244, opened 2020-11-12
monitor_adb.sh: print more verbose info for lsusb and fastboot devices

liuyq:
for debugging the adb connection lost problem

### DIFF
--- a/automated/lib/android-test-lib
+++ b/automated/lib/android-test-lib
@@ -35,7 +35,7 @@ adb_debug_info() {
 initialize_adb() {
     adb_debug_info
     adb start-server
-    timeout 600 adb wait-for-device || error_fatal "Device NOT found!"
+    timeout 300 adb wait-for-device || error_fatal "Device NOT found!"
     adb devices
 
     if [ -z "${ANDROID_SERIAL}" ]; then
--- a/automated/lib/android-test-lib
+++ b/automated/lib/android-test-lib
@@ -62,7 +62,12 @@ adb_root() {
     if [ "$(adb shell whoami)" = "root" ]; then
         echo "DUT already has adbd running as root"
     else
+        echo "lsusb -v before adb root"
+        lsusb -v
         adb root || info_msg "adb root timed out"
+        sleep 10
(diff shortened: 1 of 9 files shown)


automated/lib/android-test-lib
```diff
@@ -62,7 +62,11 @@ adb_root() {
     if [ "$(adb shell whoami)" = "root" ]; then
         echo "DUT already has adbd running as root"
     else
+        echo "lsusb -v before adb root"
+        lsusb -v
         adb root || info_msg "adb root timed out"
+        echo "lsusb -v after adb root"
```
mwasilew, line 68: is adb root a blocking call? If not, this might be a bit too fast. We have 'wait-for-device' in the next line which suggests 'adb root' isn't blocking.

liuyq: adb root should not be blocking, but it seems caused the device number changes in the output of lsusb -v, so I added this debug for check.

mwasilew: if you're too quick, the device won't show up in lsusb -v. IIRC 'adb root' causes adb disconnect and reconnect, so the enumeration change is expected.

liuyq: adb root won't be blocking , but it seems cause the device number in the output of lsusb -v changed, so I added it here for debug.
regarding to putting it before wait-for-device, my concern is that if the wait-for-device timed out, then lsusb won't get executed so I put it just before the wait-for-device command, but as your concern on the "too fast" problem, it's possible. let me add sleep 10 before running the lsusb -v command.

mwasilew:
Could you rebase and fix the shellcheck warnings?

liuyq:
> Could you rebase and fix the shellcheck warnings?

hmm, why the changes on self-test branch is here too?
Let me close this for now, I will submit a separate change when necessary

liuyq:
> > Could you rebase and fix the shellcheck warnings?
> 
> hmm, why the changes on self-test branch is here too?
I did the wrong operation locally, I pushed to the master branch :(

PR #248, opened 2020-11-25
ml-bench

K1504296:
Creating files for ml-bench, an implementation of various ml frameworks to be deployed to devices in the lab.


mwasilew:
Could you rebase and squash? It looks like a good candidate for a single commit patch.

K1504296:
Hey Milosz, It should be all in order now but if you could give it a quick look over to make sure everything is intact I'd appreciate it.

mwasilew:
Sth went wrong I think. This merge should not be there. This makes the latter commit including a file that doesn't belong there.

PR #257, opened 2020-12-10
automated: linux: fwts: filter out x86 tests on other archs

roxell:


nareshkamboju:
There are long list of unsupported / unavailable tests on arm64.
I would like to opt for default test runs - which means  the parameter TESTS: ' ' to be passed from lava test plans.

mwasilew:
@roxell Could you add a comment with the list of possible tests?

roxell:
@mwasilew thats decided by the test suite. since its a lot of arch dependent tests the testsuite can and should decide which tests to run.

nareshkamboju:
Thanks for the patch.
Lets test suite decide what list of tests it should run.

x86_64 running 1500 Tests
arm64 running 17 tests.

However, I like the default test run by fwts.

+1

nareshkamboju (approved):

PR #259, opened 2020-12-15
automated: linux: device-tree: refactor the test

fboudra:
Prepare for adding a skip list. Make the test easier to read and follow.
* there's a couple of functions that aren't necessary.
* being root isn't necessary.
* added SPDX license identifier.

Signed-off-by: Fathi Boudra <fathi.boudra@linaro.org>

### DIFF
--- a/automated/linux/device-tree/device-tree.sh
+++ b/automated/linux/device-tree/device-tree.sh
@@ -1,22 +1,8 @@
 #!/bin/sh
+# SPDX-License-Identifier: GPL-2.0-only
+# Copyright (C) 2016-2020 Linaro Ltd.
 #
 # Device Tree test cases
-#
-# Copyright (C) 2016, Linaro Limited.
-#
-# This program is free software; you can redistribute it and/or
-# modify it under the terms of the GNU General Public License
-# as published by the Free Software Foundation; either version 2
-# of the License, or (at your option) any later version.
-#
-# This program is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-# GNU General Public License for more details.
-#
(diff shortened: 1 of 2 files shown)


fboudra:
the license check fails but the identifier exists: https://spdx.org/licenses/GPL-2.0-or-later.html

It seems the "fsfe/reuse lint" test may need some fixing.

automated/linux/device-tree/device-tree.sh
```diff
@@ -1,75 +1,47 @@
 #!/bin/sh
+# SPDX-License-Identifier: GPL-2.0-or-later
+# Copyright (C) 2016-2020 Linaro Ltd.
```
mwasilew, line 3: all files in this repository are GPL2-only by default: https://github.com/Linaro/test-definitions/blob/master/COPYRIGHTS#L14. So this bit isn't required IMHO. If you would like to change the license, you need to add GPL-2.0-or-later to LICENSES directory.

mwasilew (changes requested):

automated/linux/device-tree/device-tree.sh
fboudra: I only replaced the license text by the SPDX identifier. The license on this file is GPL2+. I'm fine to relicense the code to GPL2 only but it requires its authors to agree. So adding GPL-2.0-or-later to LICENSE doesn't mean change the license, but it means keep the file licensed GPL2+.

mwasilew: Either fix will do. I'm OK with relicensing to GPL2 (IIRC I asked authors a long time ago and there was no objections) or keeping GPL2+. License check needs the license file in the repository, that's why it's failing. You can also list it in a different section in COPYRIGHTS file without adding SPDX header.

fboudra: sounds good. I'll relicense to gpl2 only then. Thanks.

PR #260, opened 2020-12-17
noninteractive-tradefed: add support for docker method

liuyq:
Including the two change for where the boot image is saved inside the docker and workaround for the dmesg permission problem


liuyq:
Tested with jobs here, though both failed with the "tradefed - adb device lost" problem.
https://validation.linaro.org/scheduler/job/2279871#L4102
https://validation.linaro.org/scheduler/job/2279872#L8338

PR #264, opened 2021-01-05
ltp: allow to set LTP_INSTALL_PATH

ivoire:


roxell:
LGTM, the shellcheck fails due to another reason that should be fixed with https://github.com/Linaro/test-definitions/pull/265

PR #271, opened 2021-02-12
workload-automation3: Fix test_case names

roxell:
The send-to-lava.sh script expects a single word as a test case name
followed by the result. If the test case name contains multiple words
the script assumes that the second word is the test result. Let's fix
the test cases to be in the expected format to prevent errors like
this:

> echo  AEP config installed pass
> ...
> ../../utils/send-to-lava.sh ./output/result.txt
> <LAVA_SIGNAL_TESTCASE TEST_CASE_ID=AEP RESULT=config>
> Received signal: <TESTCASE> TEST_CASE_ID=AEP RESULT=config
> Bad test result: config


nareshkamboju (commented):
+1

PR #283, opened 2021-03-19
packetdrill: Fix skip install section

nareshkamboju:
When packetdrill is pre-installed on the rootfs we do not have
to install pre-required packages and build the packetdrill.
so move this section when SKIP_INSTALL is not true.

### DIFF
--- a/automated/linux/packetdrill/packetdrill.sh
+++ b/automated/linux/packetdrill/packetdrill.sh
@@ -11,12 +11,13 @@ TMP_LOG="${OUTPUT}/tmp_log.txt"
 TEST_PROGRAM=packetdrill
 TEST_PROG_VERSION=
 TEST_GIT_URL=https://github.com/google/packetdrill.git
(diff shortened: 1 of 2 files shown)


roxell:
if I want to try the latest packetdrill from git now that wont be possible without trying to install packages too.
is that what we wont?

nareshkamboju:
When you want to build and install from packetdrill repo. The install_deps will be called always in this case If the required packages already there then the installer like apt or yum will install not those packages because we are not doing force install.

nareshkamboju:
This patch work is in progress. Please wait for the merge.

nareshkamboju:
@roxell The review comments are addressed in this patch.
This is ready to merge.
Tested on nxp LAVA Lab with two scenario 

1)
Pre-build packetdrill LAVA job,
https://lavalab.nxp.com/results/333492

2)
native packerdrill LAVA job,
https://lavalab.nxp.com/results/333489

roxell:
LGTM thanks

PR #285, opened 2021-04-01
linux/aep-pre-post: Switch to python3 environment for LISA

djakov:
Convert the LISA environment to python3, as python2 is no longer
supported and it would be nice to align this also with other tools
and libraries that have already moved to python3.


roxell:
LGTM thanks

PR #286, opened 2021-04-01
workload-automation3: Save the lava-target-ip to hosts

djakov:
Save the IPv4 address of the target to /etc/hosts to make it easier
for scripts to connect to the target by using just a fixed hostname.


roxell:
LGTM thanks

PR #288, opened 2021-04-28
automated: utils: test-runner: fix black formatting issues

roxell:
With black-21.4b1-py3-none-any.whl the following black formatting issue shows up
with a diff like this:
-    """ Run a shell command. If target is specified, ssh to the given target first. """
+    """Run a shell command. If target is specified, ssh to the given target first."""


jscook2345:
LGTM

PR #291, opened 2021-05-04
automated: linux: add mmtests benchmarks

mirco-romagnoli:
MMTests is a configurable test suite that runs performance tests
against arbitrary workloads.

This commit adds to the test set the mmtests suite and exports
the test results in a JSON file.

Signed-off-by: Mirco Romagnoli <romagnoli.mirco@gmail.com>

### DIFF
--- a/automated/linux/mmtests/mmtests.sh
+++ b/automated/linux/mmtests/mmtests.sh
@@ -1,0 +1,158 @@
+#!/bin/bash
+# Sysbench is a modular, cross-platform and multi-threaded benchmark tool.
+# This runs the CPU microbench which calculates primes up to a
+# certain value using varying numbers of instances.
+# In this configuration, it calculates primes up to 25000
+# using between 1 and 2*NUMCPUS threads.
+
+# shellcheck disable=SC1091
+. ../../lib/sh-test-lib
+
+OUTPUT="$(pwd)/output"
+TEST_PROGRAM="mmtests"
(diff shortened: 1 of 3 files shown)


automated/linux/mmtests/mmtests.sh
```diff
@@ -0,0 +1,151 @@
...
+	extracted_json="../${MMTESTS_TYPE_NAME}.json"
+	./run-mmtests.sh --no-monitor --config ${MMTESTS_CONFIG_FILE} benchmark
+	./bin/extract-mmtests.pl -d work/log/ -b ${MMTESTS_TYPE_NAME} -n benchmark --print-json > $extracted_json
+}
+
+! check_root && error_msg "Please run this script as root."
+
+get_test_program "${TEST_GIT_URL}" "${TEST_DIR}" "${TEST_PROG_VERSION}" "${TEST_PROGRAM}"
+
+# Test installation.
```
roxell, line 143: This if-else block must be before the function 'get_test_program()'

mirco-romagnoli: Doing that would break the while loop at line 126 which is needed to download correctly the test source files (sometimes the mirror replies with a 404 code instead of sending the file) and, in a way, is part of the installation process.
I can put that part of code in a different function that can be called later, is that ok?

roxell:
https://lkft.validation.linaro.org/scheduler/job/2627776

PR #292, opened 2021-05-10
Fix artifacts publishing with busybox

chase-qi:
Busybox find command with argument '-regex' doesn't work, use 'grep -E'
instead for pattern filtering.


chase-qi:
@stevanradakovic this is the reason why it didn't work with allscenarios qemu image.

stevanradakovic:
+1 LGTM

PR #299, opened 2021-06-14
automated: linux: add lescan to hci-smoke

mwasilew:

### DIFF
--- a/automated/linux/hci-smoke/hci-smoke-test.sh
+++ b/automated/linux/hci-smoke/hci-smoke-test.sh
@@ -3,6 +3,7 @@
 # HCI smoke tests
 #
 # Copyright (C) 2017, Linaro Limited.
+# Copyright (C) 2021, Foundries.io.
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
@@ -19,6 +20,7 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 #
 # Author: Nicolas Dechesne <nicolas.dechesne@linaro.org>
+# Author: Milosz Wasilewski <milosz.wasilewski@foundries.io>
 #
 
 # shellcheck disable=SC1091
@@ -28,19 +30,21 @@ RESULT_FILE="${OUTPUT}/result.txt"
 export RESULT_FILE
 DEVICE="hci0"
 BOOT="enabled"
+MODE="scan"
 
 # Pattern for HCI device baddr
 BADDR_PATTERN="..:..:..:..:..:.."
 
 usage() {
-    echo "Usage: $0 [-b <enabled|disabled>] [-d <device>]" 1>&2
+    echo "Usage: $0 [-b <enabled|disabled>] [-d <device>] [-m <scan|lescan|scan lescan>]" 1>&2
     exit 1
 }
 
-while getopts "d:b:" o; do
+while getopts "d:b:m:" o; do
   case "$o" in
     d) DEVICE="${OPTARG}" ;;
     b) BOOT="${OPTARG}" ;;
+    m) MODE="${OPTARG}" ;;
     *) usage ;;
   esac
 done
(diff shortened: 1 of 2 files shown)


mwasilew:
@ndechesne I added lescan to the hci-smoke. The default is not to use it so current users won't be affected.

automated/linux/hci-smoke/hci-smoke.yaml
```diff
@@ -13,14 +20,19 @@ metadata:
...
+        - imx8mmevk
+        - imx6ullevk
+        - rpi4
 
 params:
     DEVICE: hci0
     # expected state on first boot: <enabled> or <disabled> or <auto>
     BOOT: enabled
+    # MODE can be set to "normal", "le" or "both"
+    MODE: normal
```
ndechesne, line 32: I think it would be just simpler if MODE includes 'scan' or 'lescan' instead of normal, le, both. And MODE should be SCAN instead.

automated/linux/hci-smoke/hci-smoke-test.sh
```diff
@@ -123,7 +135,17 @@ test_hciconfig
 test_hciconfig_boot
 test_hciconfig_down
 test_hciconfig_up
-test_hcitool_scan
+if [ "${MODE}" = "both" ] || [ "${MODE}" = "normal" ]; then
+    test_hcitool_scan
+else
+    report_skip "hciconfig-scan"
```
ndechesne, line 141: is skip really the right status? looks like the user decided not to even execute the test, so it's not 'skipped' but simply not executed. so it should not show up in the test results. I think skipped would imply that we wanted to run it, but because of 'something' we couldn't run it. no?

automated/linux/hci-smoke/hci-smoke.yaml
mwasilew: Do you ask to remove 'both' or just rename 'normal' to 'scan'?

automated/linux/hci-smoke/hci-smoke-test.sh
mwasilew: I prefer to have all possible tests listed always (for easier accounting) but I'm OK removing this part if you think it's a bad idea.

automated/linux/hci-smoke/hci-smoke.yaml
ndechesne: yes, remove 'both', and have 'scan' and 'lescan' as the 2 supported modes. Then the tester can use MODE='scan' or MODE='scan lescan'

automated/linux/hci-smoke/hci-smoke-test.sh
ndechesne: well.. you know that better than me ;) i found it odd, but if this is how we typically do it, then it's fine for me!

automated/linux/hci-smoke/hci-smoke.yaml
mwasilew: OK

automated/linux/hci-smoke/hci-smoke-test.sh
```diff
@@ -28,19 +30,21 @@ RESULT_FILE="${OUTPUT}/result.txt"
...
 DEVICE="hci0"
 BOOT="enabled"
+MODE="scan"
 
 # Pattern for HCI device baddr
 BADDR_PATTERN="..:..:..:..:..:.."
 
 usage() {
-    echo "Usage: $0 [-b <enabled|disabled>] [-d <device>]" 1>&2
+    echo "Usage: $0 [-b <enabled|disabled>] [-d <device>] [-m <normal|le|both>]" 1>&2
```
ndechesne, line 39: this needs to be updated here.

ndechesne (commented):
minor nit. otherwise looks good to me!

automated/linux/hci-smoke/hci-smoke-test.sh
mwasilew: sorry, will fix in a moment.

ndechesne (approved):

PR #301, opened 2021-06-17
automated: linux: add network-manager smoke test

mwasilew:


nareshkamboju:
@mwasilew Please share LAVA job for this test.

mwasilew:
Here you go: https://lava.infra.foundries.io/scheduler/job/744#L2224

nareshkamboju:
@mwasilew 
 which nmcli || error_msg "'nmcli' not found, exiting..."

When nmcli not found the rootfs let this exit ?

automated/linux/network-manager/network-manager.sh
```diff
@@ -0,0 +1,93 @@
...
+check_return "nmcli-general-status"
+
+nmcli -c no general hostname
+check_return "nmcli-general-hostname"
+
+# check networking status
+nmcli -c no networking connectivity | grep full
+check_return "nmcli-initial-full-connectivity"
+
+nmcli -c no networking off
```
nareshkamboju, line 61: Will it stall on NFS mounted devices ?

mwasilew:
> @mwasilew
> which nmcli || error_msg "'nmcli' not found, exiting..."
> 
> When nmcli not found the rootfs let this exit ?

yes, that's nmcli test, so it won't work without nmcli.

automated/linux/network-manager/network-manager.sh
mwasilew: yeah, it might. I didn't test that.

roxell:
I think @nareshkamboju  means that we should fail early with an err_msg... if we don't have nmcli installed.
As I read it now we will just report a failure and move to the next nmcli test and fail there too.

automated/linux/network-manager/network-manager.yaml
```diff
@@ -0,0 +1,27 @@
...
+        - functional
+    environment:
+        - lava-test-shell
+    devices:
+        - hi6220-hikey
+        - apq8016-sbc
+        - imx8mmevk
+        - imx6ullevk
+params:
+    INTERFACE: "eth0"
```
roxell, line 22: does it make sense to add skip_install ?

mwasilew: added. Sorry for delay. I took a few days off.

PR #302, opened 2021-06-23
automated/linux: add lkp-tests

Magnani95:
Use the lkp-tests utility to run integrated and custom benchmarks.

lkp-tests(Linux Kernel Performance) integrates 80+ open source test suites
and provides a standard interface to do installation, execution and result
analysis. These tests are included inside lkp repository so they can be used
with 'lkp.sh' script.
Custom tests also can be added inside lkp-suite framework.

Add a bash-script 'lkp.sh' that allows lkp-Lava interaction:
        - Manage LKP framework and single tests dependencies.
          Can skip LKP and tests installation with two different options, so
          can optimize installation time.
        - Can set LKP-git-repository, repo-version(commit) and clone-to
          directory with CLI parameters, so that a specific lkp-version can be
          run.
        - Create and install atomic-jobs from LKP test definitions.
          By default run all atomic-jobs inside a lkp-test, but can specify
          a sub-set.
        - Parse atomic-jobs output to retrieve their exit_status and report
          it correctly. There is no common procedure to get test return-value,
          since each test has its own output format.
          Main tests are reported 'pass' when all child atomic-jobs were run
          despite their individual exit status.
          'report_atomic_job()' function is used and should be modified to
          parse atomic-job output and tell what has to be reported to the
          script.
          'skip' is reported for tests which this script doesn't know about,
          even if they were executed correctly.
        - Create 'result.txt' file according to Lava specification;
          LKP output files are grouped by test, atomic-jobs(settings)
          and run-number and copied inside the output folder.

The patch focuses on 'sysbench-cpu' test. Other tests can work anyway, but
lkp.sh has to be modified inside 'report_atomic_job()', otherwise 'skip' is
reported.

Tests involved:
        - Ubuntu 20.10 on:
                - x86_64
        - Debian 10 "buster" on:
                - x86_64
                - aarch64

'lkp.yaml' run 'sysbench-cpu' test and its only atomic-job within
test-runner or Lava.

Output sample of result.txt from:
        - Default run output from lkp.yaml:

          lava-test-set start LKP
          ./sysbench-cpu-20000-100%-300s.yaml pass
          sysbench-cpu pass
          lava-test-set stop

        - 'ebizzy' single atomic-job and two from 'hachbench' to see how
          multiple-tests run are reported; neither test was coded in
          report_atomic_job so they are reported 'skip':

          lava-test-set start LKP
          ./ebizzy-10s-100x-200%.yaml skip
          ebizzy pass
          ./hackbench-socket-8-process-1600%.yaml skip
          ./hackbench-socket-8-process-50%.yaml skip
          hackbench pass
          lava-test-set stop

Signed-off-by: Andrea Magnani <magnani95@gmail.com>


automated/linux/lkp/lkp.sh
```diff
@@ -0,0 +1,347 @@
...
+	if [[ ! -e "${TEST_DIR}/working/result" ]] ; then
+		info_msg "${1} didn't produce any results."
+		REPORT_CODE="1"
+	else	# gather atomic_job exit status accordingly test-name
+		case "${test}" in
+
+		sysbench-cpu)
+			# 'grep' return 1 if !found or 2 with error (file does not exists)
+			# 'exits_fail' appears in the output only if sysbench-cpu test failed
+			grep -r "exit_fail" "${TEST_DIR}/working/result" >/dev/null
```
chase-qi, line 244: How is the `result` file generated? Can we rely on the exit code from `lkp run test`?  Maintaining test specific result parsing here sounds too much. 

IMHO, lkp.sh should be just a wrapper of a set of lkp commands. The current patch looks good for this part. For sending lkp results to lava, I think it will be ideal if we can mange to do the below two steps.

1) generate overall result for a test run by checking `lkp run test` exist code
2) convert lkp result(e.g. /lkp/result/sysbench-cpu/20000-100%-300s/82cd3896aa8e/debian/defconfig/gcc-8/5.10.0-1029-oem/0/sysbench-cpu.json) to `result.txt` for sending to lava.

Magnani95: > How is the `result` file generated? 

`result` is a simlink generated by lkp at the end of the run and links to the last result-folder.
Alternatively there is a lkp built-in command: `lkp result <test>`

> Can we rely on the exit code from `lkp run test`? Maintaining test specific result parsing here sounds too much.

> 1. generate overall result for a test run by checking `lkp run test` exist code

`lkp run <test>` exit code is not useful, because it refers to lkp suite command and not about `<test>`.
You can test it simply running this:
```
git clone https://github.com/intel/lkp-tests.git
cd lkp-tests
sudo make install
sudo lkp install
sudo lkp split-job jobs/sysbench-cpu.yaml
sudo lkp run ./sysbench-cpu-20000-100%-300s.yaml
echo $?
```
Since `lkp install ./sysbench-cpu-20000-100%-300s.yaml` wasn't run, the test should be fail because of incomplete dependencies 

> IMHO, lkp.sh should be just a wrapper of a set of lkp commands. The current patch looks good for this part. For sending lkp results to lava, I think it will be ideal if we can mange to do the below two steps.

> 2. convert lkp result(e.g. /lkp/result/sysbench-cpu/20000-100%-300s/82cd3896aa8e/debian/defconfig/gcc-8/5.10.0-1029-oem/0/sysbench-cpu.json) to `result.txt` for sending to lava.

As I've understood, the actual `<test>` (named `atomic job` by lkp-docs) exit status need to be retrieved from output results, which are different from test to test. I tried a quick "fail-run" with ebizzy and hackbench tests, but each of them need a individual and specific parsing of the output folder to find out exit status. 
Actually the output is parsed only to retrieve exit_status, so no metrics is searched for nor is sended to Lava.
But surely this function can be moved to another file/script to make everything more modular if you think so.

chase-qi: I just tested. You are right about `lkp run test` exist code. The first point here is a generic way should be used for getting exist code. If we are grepping `exit_fail` in lkp sub-test result folder, why this is different by test/sub-test? I understand the path to sub-test result folder can be retrieved dynamically using `lkp result test`, right? Maybe, the problem here is that it may return more than one paths, but we should be able work around this by removing `/lkp/result/test' folder before each sub-test run as results should be saved in lava already.

The second point here is that test metrics should be saved in lava if it is generated. Looks like it is possible to construct path to metric file as well. Something like `$(lkp result test-case-name)/test-case-name.json`

Magnani95: test test test

Magnani95: > I just tested. You are right about lkp run test exist code. The first point here is a generic way should be used for getting exist code. If we are grepping exit_fail in lkp sub-test result folder, why this is different by test/sub-test? 

A result path example is: `/lkp/result/sysbench-cpu/20000-100%-300s/debian/debian/defconfig/gcc-8/4.19.0-16-arm64/`
with `sysbench_cpu` as test, `20000-100%-300s` is <atomic-job> (test configuration).
Inside that folder there are standard lkp ouput file: `avg.csv, avg.json, desc.yaml, matrix.csv, matrix.json, stddev.csv, stddev.json` which should be comparable output with different tests run.
There is also at least one folder named with an increasing progressive integer (eg:  `0`, `1`, ...) which contains the specific test output.
From what I've saw from running `hackbench, ebizzy` and `sysbench-cpu` (successful and failed run) and comparing them, there is no common file or line to search for to understand exit status. Or at least, I've not understood there is one.

>  I understand the path to sub-test result folder can be retrieved dynamically using `lkp result test`, right? Maybe, the problem here is that it may return more than one paths, but we should be able work around this by removing `/lkp/result/test' folder before each sub-test run as results should be saved in lava already.

You are right with `lkp result test` which could return more than one path, but I never use it:
[here](https://github.com/Linaro/test-definitions/blob/2e6ed940be8708a6d33fd0ea8cb4401894fd587b/automated/linux/lkp/lkp.sh#L244) you can see that I used `result` inside `working directory`: this is a link created by lkp in the current directory;
[after](https://github.com/Linaro/test-definitions/blob/2e6ed940be8708a6d33fd0ea8cb4401894fd587b/automated/linux/lkp/lkp.sh#L277) I delete the link to avoid a successive failed run can access previous results. A successful run would overwrite the link and any problem shouldn't occur, but an incomplete run could not overwrite and trying to get old results.

> The second point here is that test metrics should be saved in lava if it is generated. Looks like it is possible to construct path to metric file as well. Something like $(lkp result test-case-name)/test-case-name.json

My job is related to an University exam and my goal is to write this wrapper, so I'm quite agnostic about test results and metrics; to retrieve the latter, some more time is required because I need to study 'sysbench-cpu' more. It could be useful if @roxell or @nareshkamboju tell me about what they want to know from the test.

roxell: > > The second point here is that test metrics should be saved in lava if it is generated. Looks like it is possible to construct path to metric file as well. Something like $(lkp result test-case-name)/test-case-name.json
> 
> My job is related to an University exam and my goal is to write this wrapper, so I'm quite agnostic about test results and metrics; to retrieve the latter, some more time is required because I need to study 'sysbench-cpu' more. It could be useful if @roxell or @nareshkamboju tell me about what they want to know from the test.

We shouldn't need to study the 'sysbench-cpu' more, the outputs should be generated in a json file. In the directory that @chase-qi shows above you can dig out the results via 'lkp result test-case-name'.
So we should just need to parse that file so LAVA understands it.

Magnani95: Oh yes, sorry I misunderstood output. A couple of day and I'll deliver the new patch. 

Thank you for your patience.

Magnani95: Hi @roxell and @chase-qi,
I rewrite "report_atomic_job" function and now it retrieves and parse main <test>.json correctly.
Now everything is also reported to result.txt file so that Lava should use test data correctly.
I hope that now the patch is fine.

chase-qi: Thanks @Magnani95. It looks good to me now.

roxell:
Thank you @Magnani95 for the contribution.

PR #311, opened 2021-09-02
Add rt-test JSON feature

igaw:
This PR updates the rt-tests test suite to version v2.2. The highlight of the new rt-test release is the support for writing the final result into a JSON formatted file. This also includes the histogram data for cyclictest.

Until now, the final results had to parsed from the output (which was tricky due to ASCII control chars for cursor movements). With JSON output this makes the parsing very simple. Also this PR adds support to upload the final JSON result to the artifactory for later parsing, e.g. creating a histogram:

image

Obviously, with this change all root file systems need to update their rt-tests suite or fall back using the static build ones provided in this tree.

Changes:
  - Parser update to read results from JSON result file
  - Add support for artifactory to upload the JSON result file.
  - rt-tests suite binary updated to v2.2
  - Make all tests run quiet (not spamming the logs anymore!)
  - A few small fixes (e.g. do not premature exist script aka don't use the evil '-e' script option)

I've been using successfully these patches for a quite a while. 

Here and example for cyclictest JSON output (please ignore the version number. it's a bug):

```JSON
{
  "file_version": 1,
  "cmdline:": "/usr/bin/cyclictest -q -p 98 -i 1000 -t 2 -a 0-1 -D 1m -h 100 -m --json=/lava-2950/0/tests/0_cyclictest/automated/linux/cyclictest/output/cyclictest.json",
  "rt_test_version:": "2.20",
  "start_time": "Thu, 02 Sep 2021 21:34:49 +0200",
  "end_time": "Thu, 02 Sep 2021 21:35:49 +0200",
  "return_code": 0,
  "sysinfo": {
    "sysname": "Linux",
    "nodename": "lava",
    "release": "5.14.0-rt16",
    "version": "#4 SMP PREEMPT_RT Thu Sep 2 21:29:38 CEST 2021",
    "machine": "x86_64",
    "realtime": 1
  },
  "num_threads": 2,
  "resolution_in_ns": 0,
  "thread": {
    "0": {
      "histogram": {
        "1": 24029,
        "2": 12297,
        "3": 21492,
        "4": 2081,
        "5": 95,
        "6": 6
      },
      "cycles": 60000,
      "min": 1,
      "max": 6,
      "avg": 2.03,
      "cpu": 0,
      "node": 0
    },
    "1": {
      "histogram": {
        "1": 23386,
        "2": 12292,
        "3": 21377,
        "4": 2780,
        "5": 144,
        "6": 12,
        "7": 1
      },
      "cycles": 59992,
      "min": 1,
      "max": 7,
      "avg": 2.07,
      "cpu": 1,
      "node": 0
    }
  }
}
```


igaw:
FYI, @khilman another update for rt-tests.

nareshkamboju (approved):

nareshkamboju (approved):

automated/linux/pi-stress/pi-stress.sh
```diff
@@ -53,20 +53,13 @@ fi
 
 background_process_start bgcmd --cmd "${BACKGROUND_CMD}"
 
-# pi_stress will send SIGTERM when test fails. The single will terminate the
+# pi_stress will send SIGTERM when test fails. The signal will terminate the
 # test script. Catch and ignore it with trap.
 trap '' TERM
-"${binary}" --duration "${DURATION}" "${MLOCKALL}" "${RR}" | tee "${LOGFILE}"
+"${binary}" -q --duration "${DURATION}" ${MLOCKALL} ${RR} --json="${LOGFILE}"
```
bhcopeland, line 59: Shouldn't ${MLOCKALL} ${RR} be escaped?

igaw: Yes, let me update it.

igaw: Ah no wait. The very purpose of this commit is to drop the quotes. I tried to express this in the commit message but obviously failed.  When the user sets MLOCKALL to false (or RR to false) the resulting command line contains a two apostrophe which get translated into an empty argv entry. This is makes the command line parser stop processing the command line and hence we will miss any command line option after the first empty quote. Just to illustrate it, in case you
set MLOCKALL=false and RR=true, the RR=true would never be picked up.

Since we have two command line option which could be set to false, we can't just move everything non 'false' before the last command line option. I am no bash wizard, so there might be another way to fix this problem. I just figured dropping
the quotes makes the parser work.

bhcopeland (approved):

automated/linux/pi-stress/pi-stress.sh
roxell: make sense, thank you for the explanation

verschwindibussen:
Thanks!

PR #321, opened 2021-11-18
test-runner: Add support for reporting to fiotest

mwasilew:
fiotest is a test runner/executor that works on Foundries.io LmP in a
docker container. More details in the repository:
https://github.com/foundriesio/fiotest

Signed-off-by: Milosz Wasilewski <milosz.wasilewski@foundries.io>

PR #322, opened 2021-11-18
Kernel config lib

mwasilew:
This PR fixes issue with running IMA test with test-runner on a remote target. It contains the commit submitted as #321.


mwasilew:
@roxell @chase-qi please take a look

automated/utils/test-runner.py
```diff
@@ -907,6 +909,42 @@ def send_to_lava(self, data):
...
         subprocess.call(shlex.split(cmd))
 
+    def send_to_fiotest(self):
+        """
+        This method saves results as filesystem tree. This is required by
+        fiotest: https://github.com/foundriesio/fiotest/
+        """
+        # check if TEST_DIR variable is set
+        test_path = os.environ.get("TEST_DIR")
+        if not test_path:
```
chase-qi, line 919: @mwasilew do you think it make sense to return here directly if test_path not set? When the path is not set(in most cases), printing the following two warning lines for every test probably is too much.

```
2021-11-23 14:54:24,216 - RUNNER.ResultParser: WARNING: TEST_DIR is not set
2021-11-23 14:54:24,216 - RUNNER.ResultParser: WARNING: NOT reporting result to fiotest
```

mwasilew: Fine by me, I'll update the patch - change to debug

mwasilew:
Looks like the other patch was approved. I'll rebase and resubmit.

chase-qi:
Thanks, Milosz. It looks good me @roxell

PR #325, opened 2021-11-29
automated: add support for udhcpc in wlan-download test

mwasilew:
This patch fixes the issue with wlan-download test when it's used on
busybox like environment. When dhclient is not available there is a
chance it can be replaced with udhcpc.


mwasilew:
@alimon @roxell @chase-qi could you take a look? I needed to run this test in busybox and made some changes.

roxell:
LGTM

nareshkamboju (approved):
Thanks for the patch.

PR #326, opened 2021-12-03
Use archived tree in Android ART microbenchmarks instead of pulling a tree using repo.

akosdenke:
This patch allows to use the same source tree, which was used during the build part.
Before this, a new source tree were fetched from scratch, which may missing some source customization used during the build.


vishalbhoj:
@roxell @nareshkamboju  Can you guys please review this ?

liuyq:
@vishalbhoj @akosdenke it looks good to me. let's see if @roxell and @nareshkamboju have any comment on it. BTW, what's the problem with the repo method?

akosdenke:
> BTW, what's the problem with the repo method?

It was not containing the extra patches we apply at the building phase of the CI. Those patches are available on the Linaro internal Gerrit that the LAVA worker cannot reach.
Also, it seems a much cleaner setup to use exactly the same source upon executing the tests as it was used during the build. This is why we changed to use a better accessible archived version of a tree than cloning another instance of sources which can differ.

automated/android/microbenchmarks/microbenchmarks.sh
```diff
@@ -51,23 +51,23 @@ fi
...
 tar -xf "${BUILD_TARBALL}"
+
+[ -z "$HOME" ] && export HOME="/"
 export PATH=${PWD}/out/host/linux-x86/bin/:${PATH}
 
 # FIXME removing latest adb from build since it is not working well from container
 rm -rf "${PWD}"/out/host/linux-x86/bin/adb
 initialize_adb
 adb_root
-curl https://storage.googleapis.com/git-repo-downloads/repo > "${PWD}"/out/host/linux-x86/bin/repo
```
roxell, line 66: Can't we remove 'curl' from PKG_DEPS, since that isn't used anymore what I can see.

akosdenke: Good point!
I will do in the next commit!

akosdenke:
> Please add a more descriptive changelog.

Added additional text to the initial commit.
If you meant something else under 'changelog' than please clarify.

roxell:
> > Please add a more descriptive changelog.
> 
> Added additional text to the initial comment. If you meant something else under 'changelog' than please clarify.

Yes, I meant so its part of the patch commit changelog.

Also could you provide a link to the SNAPSHOTS_URL ?

roxell:
Thank you for the update @akosdenke , can you please add your Signed-off-by as well.

akosdenke:
> Thank you for the update @akosdenke , can you please add your Signed-off-by as well.

Done! Can you please rather show me some document about the contribution requirements? It is not the best to get these information one-by-one from you. I assume there is document officially states these which I can follow next time.

akosdenke:
Also, I still owe you a snapshot URL:
https://snapshots.linaro.org/android/linaro-art-tip-build-MicroBenchmarks/584/
It is protected though, but I can ensure you that I tried it already by submitting a LAVA job pointing to my forked repo and it produced the same results as the previous jobs.
You may check it here, but it is also protected:
https://validation.linaro.org/scheduler/job/2907239/definition

PR #330, opened 2021-12-13
automated: fix IMA test to avoid using dmesg

mwasilew:
When IMA test is executed a long time after the boot, the initialization
message might already be gone from dmesg buffer. In this case the test
will falsly report 'fail' result. This patch fixes the issue by checking
if the actual policy file exists on the sysfs.


nareshkamboju (approved):

mwasilew:
Thanks @nareshkamboju!

PR #332, opened 2022-01-05
automated: linux: add timesyncd NTP sync test

mwasilew:
This test checks whether timesyncd is running and synchronizing to NTP
server. As additional check reference NTP server's time is compared with
local system time.

### DIFF
--- a/automated/linux/timesyncd/timesynctest.sh
+++ b/automated/linux/timesyncd/timesynctest.sh
@@ -1,0 +1,66 @@
+#!/bin/sh
+#
+# systemd-timesyncd test
+#
+# Copyright (C) 2021, Foundries.io
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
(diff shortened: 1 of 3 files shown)


automated/linux/timesyncd/timesynctest.sh
```diff
@@ -0,0 +1,64 @@
...
+        h|*) usage ;;
+    esac
+done
+
+! check_root && error_msg "You need to be root to run this script."
+
+if [ "$SKIP_INSTALL" = 'true' ] || [ "$SKIP_INSTALL" = 'True' ]; then
+    warn_msg "Dependencies for timesyncd test installation skipped!"
+else
+    install_deps "python3"
```
roxell, line 48: shouldn't this be "python3-pip"?

automated/linux/timesyncd/timesynctest.sh
```diff
@@ -0,0 +1,64 @@
...
+
+timedatectl show -p CanNTP --value | grep "yes"
+check_return "timesyncd-canntp"
+timedatectl show -p NTP --value | grep "yes"
+check_return "timesyncd-ntp"
+timedatectl show -p NTPSynchronized --value | grep "yes"
+check_return "timesyncd-ntpsynchronized"
+
+# check if local time is the same a served by NTP
+python3 -m pip install --user ntplib || true
```
roxell, line 62: shouldn't we just fail if we can't install ntplib ?

mwasilew: indeed, but it's easier to fail in the python script - less shell code.

automated/linux/timesyncd/timesynctest.sh
mwasilew: let me double check why I did it this way. You're probably correct. I'm using pre-built OS so might miss this kind of issue

mwasilew: fixed

automated/linux/timesyncd/timesynctest.sh
roxell: right, make sense

nareshkamboju (commented):
Looks good to me.

PR #339, opened 2022-01-27
automated: linux: force connectivity check in NM test

mwasilew:
NetworkManager caches connectivity status. Sometimes this causes an
unexpected failure in the test. This patch forces NM to check for the
connectivity status every time there is a change in the interfaces.


nareshkamboju (approved):
LGTM

PR #343, opened 2022-02-04
pkcs11/se05x tests

mwasilew:
This patch adds tests that exercise pkcs11 features. They can be used together with se05x TPM devices.

### DIFF
--- a/automated/linux/pkcs11/pkcs11.sh
+++ b/automated/linux/pkcs11/pkcs11.sh
@@ -1,0 +1,341 @@
+#!/bin/bash
+# SPDX-License-Identifier: GPL-2.0-only
+# Copyright (C) 2021 Foundries.io Ltd.
+
+# shellcheck disable=SC1091
+. ../../lib/sh-test-lib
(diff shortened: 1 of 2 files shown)


nareshkamboju (approved):

automated/linux/pkcs11/pkcs11.sh
```diff
@@ -0,0 +1,341 @@
...
+        s) SKIP_INSTALL="${OPTARG}";;
+        l) EXECUTE_LOOP="${OPTARG}";;
+        h|*) usage ; exit 1 ;;
+    esac
+done
+
+! check_root && error_msg "You need to be root to run this script."
+# Test run.
+create_out_dir "${OUTPUT}"
+install_deps "opensc bc" "${SKIP_INSTALL}"
```
roxell, line 48: should 'openssl' be part of the install_deps?

mwasilew: I think you're right, I'll add it. The ssscli doesn't have a package in any distro (as far as I'm aware).

PR #345, opened 2022-02-15
automated: test-runner: cleanup repository copy

mwasilew:
When executing tests with test-runner a copy of test-definitions
repository is created. This copy is not removed after tests are
complete. When running repeated tests in the same environment this leads
to filling up the storage. This patch prevents the problem by removing
the copy of test-definitions along with other temporary files after
tests were executed. The downside is that all the results and logs are
removed as well. Therfore the --cleanup option should only be used when
using external test result repository (like LAVA or SQUAD). The option
is disabled by default.


roxell:
LGTM

PR #351, opened 2022-03-03
mmtests: adapt for workloads with multiple subtests

federico-gelmo:
mmtests.sh: adapt for correct generation of json files
when the workload supplied is composed by more than one benchmark,
as in the case of workload-aim9-pagealloc

json-to-lava: fix incorrect LAVA output when supplied with
operation names containing "/"

Signed-off-by: Federico Gelmetti <federico.gelmo@gmail.com>


nareshkamboju (approved):

nareshkamboju:
Please post results summary before and after your patch applied.

PR #358, opened 2022-03-28
OTA rollback debug

mwasilew:
OTA rollback test had issues with devices that use RPMB instead of u-boot environment.  Some variables are only set by aktualizr-lite when is starts. This patch fixes the problem.


roxell:
LGTM

nareshkamboju (approved):

PR #367, opened 2022-06-24
linux: ltp: update skipfile-lkft with tuxrun targets

roxell:
Tuxrun is a commandline tool for testing Linux on QEMU or FVP.


bhcopeland:
Is the - vs _ naming intentional?

roxell:
> Is the - vs _ naming intentional?

yes its intentional, since there are two different qemu's devices for now.
The lab will update their devices to to use qemu-*.

nareshkamboju (approved):

nareshkamboju:
Let the LAVA lab update the qemu names.

PR #369, opened 2022-06-30
Add Real-time Linux Analysis tool

igaw:
The rtla is a meta-tool that includes a set of commands that aims to analyze the real-time properties of Linux. But instead of testing Linux as a black box, rtla leverages kernel tracing capabilities to provide precise information about the properties and root causes of unexpected results.

- osnoise
  Gives information about the operating system noise (osnoise).

- timerlat
  Measures the IRQ and thread timer latency.

See for more information: https://docs.kernel.org/tools/rtla/rtla.html

This is an initial integration into test-definitions, following the rt-tests as mach as possible (even faking the JSON format output). Maybe upstream will add support for native JSON output but until let's role with this format.

I haven't added any access to any command line options except the duration argument for this version. The aim is to get first some sane defaults figured out.

These tests depend on following kernel options:

```
CONFIG_OSNOISE_TRACER=y
CONFIG_TIMERLAT_TRACER=y
CONFIG_HIST_TRIGGERS=y
```

image

### DIFF
--- a/automated/linux/rtla/parse_rtla.py
+++ b/automated/linux/rtla/parse_rtla.py
@@ -1,0 +1,107 @@
+#!/usr/bin/env python3
+import argparse
+import platform
+import json
+
+# Transform rtla hist out data into the same format as rt-tests
+# framework is producing.
+
+
+def parse_args():
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "-r",
+        "--result-file",
+        dest="result_file",
+        required=True,
+        default="./result.txt",
+        help="Specify test result file.",
+    )
+    parser.add_argument(
+        "-t",
+        "--test-name",
+        required=True,
+        help="Specify test name.",
+    )
+    parser.add_argument(
+        "-o",
+        "--output",
+        dest="output",
+        help="Specify output file.",
+    )
+
+    args = parser.parse_args()
+    return args
+
+
+def get_sysinfo():
+    sysinfo = {}
+
+    uname = platform.uname()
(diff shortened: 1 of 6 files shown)


automated/linux/rtla/parse_rtla.py
```diff
@@ -0,0 +1,104 @@
...
+    )
+
+    args = parser.parse_args()
+    return args
+
+
+def get_sysinfo():
+    sysinfo = {}
+    uname = platform.uname()
+    print(type(uname))
```
igaw, line 40: urgh, a debug leftover...

bristot:
why do you consider it incorrect?

igaw:
I am missing userspace comms, like hackbench when I compare it to the output of timerlat. Also 4 hits only?

bristot:
If you are running osnoise with FIFO priority, all the background noise (running as sched other) will not preempt osnoise. The only workload to preempt osnoise will be those with a priority higher than the osnoise threads, as the migration threads. So the output looks correct.

The thread noise for timerlat has a slightly different meaning. It accounts for both the high priority threads that might delay the timerlat/CPU thread because of obvious reasons (the so-called interference) and the lower priority thread that was running when the timerlat IRQ wakes up the timerlat thread (the so-called preemption delay). Thus, timerlat will always show at least one thread latency per activation (the thread running when the timerlat IRQ ran to wake up the timerlat thread).

igaw:
Thanks for explaining. Makes sense. So all good.

nareshkamboju:
Thanks for the patch.

PR #374, opened 2022-07-15
automated: linux: add libcurl SAN-less test

mwasilew:
This test covers the bug in libcurl that results in Out of memory
exception when dealing with HTTPS server with self-signed certificate.
When certificate lacks SAN fieds, libcurl would quit. Github error:
https://github.com/curl/curl/issues/8559

The bug was fixed in libcurl:
https://github.com/curl/curl/pull/8560

The fix is not propagated to all distros yet (as of July 2022).


roxell:
libcurl doesn't crash it just return the wrong fault.

Part from that I think the patch looks good.

PR #379, opened 2022-09-02
noninteractive-tradefed: add plugdev group for the testuser

liuyq:
so that it could access the adb and fastboot devices on the host
as the default user


liuyq:
tested with the job here, and it passed now:
https://validation.linaro.org/scheduler/job/3641155

PR #381, opened 2022-09-03
initialize_adb: unset ANDROID_SERIAL before get-serialno

liuyq:
when ANDROID_SERIAL is empty, to work around the following problem
    ++ adb get-serialno
    error: device '' not found
    + ANDROID_SERIAL=


liuyq:
tested with the lava job here:
https://lkft.validation.linaro.org/scheduler/job/5480841

PR #383, opened 2022-09-21
automated: linux: add ostree only option to ota-update

mwasilew:
This patch allows to test type=ostree when using aktualizr-lite for pulling OTA updates.


automated/linux/ota-update/download-update.sh
```diff
@@ -73,7 +79,12 @@ cp aklite-callback.sh /var/sota/
 chmod 755 /var/sota/aklite-callback.sh
 
 mkdir -p /etc/sota/conf.d
-cp z-99-aklite-callback.toml /etc/sota/conf.d/
+mkdir -p /etc/sota/conf.d
```
roxell, line 82: copy-paste from ota-rollback I guess, one line too much.

Would it make sense to harmonise those two tests? since a lot of the code are the same...

mwasilew: I tried, but it's harder than it seems. There are a lot of small differences. Fixed this issue

mwasilew:
I missed the commend. github doesn't send me any notifications when there are comments on PR :(

PR #385, opened 2022-10-06
Add Secure Boot Enabled definition

jetm:
Validate Secure Boot is enabled.

### DIFF
--- a/automated/linux/secure-boot/secure-boot-enabled.sh
+++ b/automated/linux/secure-boot/secure-boot-enabled.sh
@@ -1,0 +1,59 @@
+#!/bin/bash
+# SPDX-License-Identifier: GPL-2.0-only
+# Copyright (C) 2022 Linaro Ltd.
+
+# shellcheck disable=SC1091
+. ../../lib/sh-test-lib
+OUTPUT="$(pwd)/output"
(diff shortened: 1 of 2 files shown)


automated/linux/secure-boot/secure-boot-enable.yaml
```diff
@@ -0,0 +1,11 @@
...
+  format: Lava-Test Test Definition 1.0
+  name: secure-boot
+  description: "Check SecureBoot is enabled"
+  os:
+    - linux-any
+  scope:
+    - functional
+run:
+  steps:
+    - lava-test-case "secure-boot-enabled" --shell test "$(efivar -d -n 8be4df61-93ca-11d2-aa0d-00e098032b8c-SecureBoot)" -eq 1
```
mwasilew, line 11: Could you convert this to a script that takes a parameter so it's reusable?

mwasilew (changes requested):

automated/linux/secure-boot/secure-boot-enable.yaml
jetm: @mwasilew I created the script without the parameter because the variable is hard coded. I have put more detail in the description

PR #387, opened 2022-10-13
lib: sh-test-lib: Add riscv64 support

nkbelin:
Add the support of the riscv64 platform by adding the skipgen executable for riscv64.
The binary has been generated following the "Releasing" instructions at https://github.com/Linaro/skipgen using go version go1.18.1 and goreleaser. The difference is that the upx tool to compress the executable has not been used as it does support the riscv64 architecture at the moment.

Signed-off-by: Nicolas Belin <nbelin@baylibre.com>


roxell:
> @roxell it's getting big, any ideas on moving skipgen binaries out of the repository? Alternatively can we separate test-definitions-core so it is easier to create derivative repositories with only required stuff?

Moving this discussion into an issue [1], because it deserves its own thread.
That said, this shouldn't hold back this PR.
PR looks good to me.

[1] https://github.com/Linaro/test-definitions/issues/388

PR #389, opened 2022-10-24
automated/linux/measured-boot: adding measured-boot test cases

K1504296:
Creating measured boot test cases for trusted substrate check that measured boot is enabled.

measured-boot-enabled-check-PCR7.yaml checks the value in PCR7 and reboots the device.
measured-boot-enabled-badboot.yaml changes the PCR7 value and afterwards it boots the device with the bad value.

If measured boot from TS is enabled the device should not boot. If it does boot, run check-PCR7 again to make sure PCR7 value has actually changed.

### DIFF
--- a/automated/linux/measured-boot/measured-boot-enable-badboot.yaml
+++ b/automated/linux/measured-boot/measured-boot-enable-badboot.yaml
@@ -1,0 +1,29 @@
+metadata:
+    name: measured-boot-changed-PCR7
+    format: "Lava-Test Test Definition 1.0"
+    description: "Checking PCR7 value for measured boot testing"
+    maintainer:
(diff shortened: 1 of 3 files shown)


chaws:
Hi @K1504296 could you please follow guidelines here: https://github.com/Linaro/test-definitions/blob/master/docs/test-writing-guidelines.md#writing-a-test-case? For example, you could base your measure-boot test off https://github.com/Linaro/test-definitions/tree/master/automated/linux/optee, which has a shell script that not only knows how to run the test but also how to obtain it in case it doesn't exist locally.

@roxell is the main maintainer of this project and he recommends following the guidelines to ensure tests are reusable elsewhere.

PR #401, opened 2022-11-22
automated: linux: add more pkcs11 tests

mwasilew:
When running loop test in pkcs11, TA may run out of memory. This isn't a recoverable error and the remaining results should not be trusted. This patch adds a check which indicates OOM situation with a proper test result. It allows to diagnose the problem manually afterwards.

### DIFF
--- a/automated/linux/pkcs11/pkcs11.sh
+++ b/automated/linux/pkcs11/pkcs11.sh
@@ -281,17 +281,37 @@ test_rsa_loop()
 {
     se05x_connect
     BREAK="False"
+    LABEL=0
     while [ "$BREAK" = "False" ]
     do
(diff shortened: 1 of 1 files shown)


mwasilew:
@roxell has anything changed in the CI? The linter results seem to be wrong. Shellcheck ignores the `# shellcheck disable=` lines. It works well on the previous PR.

roxell:
> @roxell has anything changed in the CI? The linter results seem to be wrong. Shellcheck ignores the `# shellcheck disable=` lines. It works well on the previous PR.

looks like the new version of shellcheck 0.8.0 that is more strict.

roxell:
running shellcheck on your changed files doesn't complain on anything so I'll merge this.

Thanks

PR #406, opened 2022-11-25
automated: linux: kunit: fix parsing

roxell:
Slugify test names so LAVA can understand the format. This will work with the old kunit way
ok 1 - test_1..
and
ok 1 test_1..

### DIFF
--- a/automated/linux/kunit/parse-output.py
+++ b/automated/linux/kunit/parse-output.py
@@ -1,0 +1,29 @@
+#!/usr/bin/env python3
+import sys
+import re
+
+
+def slugify(line):
+    non_ascii_pattern = r"[^A-Za-z0-9_-]+"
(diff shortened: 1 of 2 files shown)


automated/linux/kunit/parse-output.py
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env python3
+import sys
+import re
+
+
+def slugify(line):
+    non_ascii_pattern = r"[^A-Za-z0-9]+"
```
mwasilew, line 7: I think you can add _- here as well. LAVA is OK with underscores and hyphens. It doesn't like white spaces.

roxell: done, thanks you for your input.

nareshkamboju (approved):

PR #416, opened 2023-01-02
kselftest: skipfile-lkft: Adding arm64 fp-stress to skipfile

nareshkamboju:
The kselftest arm64 fp-stress test case hangs on qemu-arm64. Due to this hang test other tests not being run.


roxell (approved):
LGTM

PR #417, opened 2023-01-02
support ssh with non-standard port number and opkg

mikkorapeli-linaro:
Enables running tests on qemu machine which exposes ssh on localhost IP address and non-standard TCP port number.

Then capture package list also with opkg used by some yocto build configurations.

### DIFF
--- a/automated/utils/test-runner.py
+++ b/automated/utils/test-runner.py
@@ -629,6 +629,9 @@ def get_packages(linux_distribution, target=None, target_port=None):
     linux_distribution is a string that may be 'debian',
         'ubuntu', 'centos', or 'fedora'.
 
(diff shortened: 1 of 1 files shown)


mikkorapeli-linaro:
Can anyone show the details of the check failure with sanity-check.sh? It's passing here with python 3.10.6 from Ubuntu 22.04 and the failure logs don't have any details.

roxell:
> Can anyone show the details of the check failure with sanity-check.sh?

its black that fails its checks
black --check automated/utils/test-runner.py
if you want black to fix the issue just do
black automated/utils/test-runner.py

mwasilew (approved):

mikkorapeli-linaro:
Not relevant anymore.

PR #421, opened 2023-01-10
Fix small parser issues

nuclearcat:
After upgrading perl parser to python few small issues was discovered. One of them is failure to parse "not ok" result, another one is likely https://github.com/Linaro/test-definitions/pull/418 require more proper fix in some conditions.
Both commits tested on staging in kernelci and collabora LAVA: https://lava.collabora.dev/scheduler/job/8642420

### DIFF
--- a/automated/linux/kselftest/kselftest.sh
+++ b/automated/linux/kselftest/kselftest.sh
@@ -142,6 +142,7 @@ install() {
 }
 
 ! check_root && error_msg "You need to be root to run this script."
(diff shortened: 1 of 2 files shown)


gctucker:
@nuclearcat Have you also made a PR for the KernelCI fork or tried this on staging.kernelci.org somehow?

nuclearcat:
Yes, this is how i got it tested: https://github.com/kernelci/test-definitions/pull/19

PR #428, opened 2023-02-15
Add docker-prune tests

brenwoo:
Tests that check whether docker correctly prunes
app images after update or disable

### DIFF
--- a/automated/linux/docker-prune/disable-prune.sh
+++ b/automated/linux/docker-prune/disable-prune.sh
@@ -1,0 +1,47 @@
+#!/bin/bash
+# shellcheck disable=SC1091
+. ../../lib/sh-test-lib
+. ./prune-lib.sh
+
+OUTPUT="$(pwd)/output"
+RESULT_FILE="${OUTPUT}/result.txt"
+export RESULT_FILE
+
+TOKEN=$(cat /etc/lmp-device-register-token)
+DEVICE_NAME=$(cat /etc/hostname)
+API_DATA='{ "reason": "Prune Test", "files":
+[{"name":"z-50-fioctl.toml","on-changed":["/usr/share/fioconfig/handlers/aktualizr-toml-update"],"unencrypted":true,"value":"\n[pacman]\n compose_apps = \"\"\n"}]}'
+FACTORY=$(grep -w "LMP_FACTORY" /etc/os-release | cut -d'=' -f2 | sed 's/\"//g')
+
+
+! check_root && error_msg "You need to be root to run this script."
+create_out_dir "${OUTPUT}"
+
+setup_callback
+
+auto_register
+
+while ! docker ps -a | grep "shellhttpd"; do
+    echo "waiting for container"
+    sleep 1
+done
+
+get_image_sha
+
+curl --data "$API_DATA" -H "Content-Type: application/json" -H "OSF-TOKEN: $TOKEN" -X PATCH https://api.foundries.io/ota/devices/"$DEVICE_NAME"/config/?factory="$FACTORY"
+
+wait_for_signal
+
+if check_image_prune; then
+    report_pass "disable-prune"
+else
+    report_fail "disable-prune"
+fi
+
+if check_image shellhttpd; then
+    report_fail "image-removed"
+else
+    report_pass "image-removed"
+fi
+
+rm "$(pwd)/sha.txt"
--- a/automated/linux/docker-prune/prune-lib.sh
+++ b/automated/linux/docker-prune/prune-lib.sh
@@ -1,0 +1,88 @@
+#!/bin/bash
+
+verify_app_present () {
+    app="$1"
+
+    if docker ps | grep -q "${app}"; then
+        return 0
+    fi
+
+    return 1
+}
+
+check_image () {
+    image_name="$1"
+
+    if [[ "$(docker images -q "${image_name}" 2> /dev/null)" == "" ]] ; then
+        return 1
+    fi
+
+    return 0
+}
+
+get_image_sha () {
+    ls /var/sota/reset-apps/apps/shellhttpd > "$(pwd)/sha.txt"
+}
+
+check_image_prune () {
+    runtime="5 minutes"
+    endtime=$(date -ud "$runtime" +%s)
+
+    while true; do
+        if [ "$(date -u +%s)" -ge "$endtime" ]; then
+            echo "Not found"
+            return 1
+        elif (journalctl --no-pager -u aktualizr-lite | grep "Pruning unused docker containers"); then
+            echo "Found"
+            return 0
+        fi
+        sleep 1
+   done
+}
+
+compare_sha () {
+    if diff "$(pwd)/sha.txt" <(ls /var/sota/reset-apps/apps/shellhttpd) > /dev/null; then
+        return 1
+    fi
+
+    return 0
+}
+
+setup_callback () {
+    cp aklite-callback.sh /var/sota/
+    chmod 755 /var/sota/aklite-callback.sh
+    mkdir -p /etc/sota/conf.d
+    cp z-99-aklite-callback.toml /etc/sota/conf.d/
+    report_pass "create-aklite-callback"
+    touch /var/sota/ota.signal
+    touch /var/sota/ota.result
+    report_pass "create-signal-files"
+}
+
+wait_for_signal () {
+    SIGNAL=$(</var/sota/ota.signal)
+    while [ ! "${SIGNAL}" = "check-for-update-post" ]
+    do
+        echo "Sleeping 1s"
+        sleep 1
+        cat /var/sota/ota.signal
+        SIGNAL=$(</var/sota/ota.signal)
+        echo "SIGNAL: ${SIGNAL}."
+    done
+    report_pass "update-post-received"
+}
+
+auto_register () {
+    if [ -f "/var/sota/sql.db" ]; then
+        echo "Device registered, skipping registration"
+    else
+        systemctl enable --now lmp-device-auto-register || error_fatal "Unable to register device"
+    fi
+
+    while ! systemctl is-active aktualizr-lite; do
+        echo "Waiting for aktualizr-lite to start"
+        sleep 1
+    done
+# add some delay so aklite can setup variables
+    sleep 5
+}
--- a/automated/linux/docker-prune/prune.sh
+++ b/automated/linux/docker-prune/prune.sh
@@ -1,0 +1,39 @@
+#!/bin/bash
+# shellcheck disable=SC1091
+. ../../lib/sh-test-lib
+. ./prune-lib.sh
+
+OUTPUT="$(pwd)/output"
+RESULT_FILE="${OUTPUT}/result.txt"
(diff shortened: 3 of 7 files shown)


automated/linux/docker-prune/disable_prune.sh
```diff
@@ -0,0 +1,65 @@
...
+    FOUND=0
+
+    while [ ! "${FOUND}" -eq 1 ]; do
+        if [ "$(date -u +%s)" -ge "$endtime" ]; then
+            break
+            return 1
+        elif (journalctl --no-pager -u aktualizr-lite | grep "Pruning unused docker containers"); then
+            echo "Prune Successful"
+            FOUND=1
+            return  0
```
roxell, line 26: do you need to set FOUND when you always return?
Can this be done in a simpler way?

automated/linux/docker-prune/disable_prune.yaml
```diff
@@ -0,0 +1,22 @@
...
+    name: docker-prune
+    description: "Prune app image after disabling app"
+
+    maintainer:
+        - brendan.wood@foundries.io
+    os:
+        - openembedded
+    scope:
+        - functional
+
```
roxell, line 14: drop newline

automated/linux/docker-prune/disable_prune.yaml
```diff
@@ -0,0 +1,22 @@
...
+    devices:
+        - imx8mm
+
+run:
+    steps:
+        - cd ./automated/linux/docker-prune
+        - ./disable_prune.sh
```
roxell, line 21: remove trailing whitespace

automated/linux/docker-prune/prune.yaml
```diff
@@ -0,0 +1,23 @@
...
+    description: "Download OTA update and prune
+    old container images"
+
+    maintainer:
+        - brendan.wood@foundries.io
+    os:
+        - openembedded
+    scope:
+        - functional
+
```
roxell, line 15: drop newline

automated/linux/docker-prune/prune.yaml
```diff
@@ -0,0 +1,23 @@
...
+    devices:
+        - imx8mm
+
+run:
+    steps:
+        - cd ./automated/linux/docker-prune
+        - ./prune.sh
```
roxell, line 22: drop trailing whitespace

brenwoo:
> Maybe, disable_prune.sh and prune.sh should be consolidated into one script or it should have a lib that both scripts can use... since its a lot that is in common. Thoughts?

Yes, a lib makes sense. They are separate scripts to make the timings easier, but a lib will remove some redundancy

automated/linux/docker-prune/disable_prune.sh
```diff
@@ -0,0 +1,65 @@
...
+    echo "${APP} container found"
+    report_pass "app-present"
+else
+    echo "${APP} container not found"
+    report_fail "app-present"
+    skip_list="run-disable-prune"
+    exit_on_fail "verify-app-present""${skip_list}"
+fi
+
+curl --data "$API_DATA" -H "Content-Type: application/json" -H "OSF-TOKEN: $TOKEN" -X PATCH https://api.foundries.io/ota/devices/"$DEVICE_NAME"/config/?factory="$FACTORY"
```
mwasilew, line 57: This doesn't look right. The .toml is not .json. Does it actually work?

automated/linux/docker-prune/prune.sh
```diff
@@ -0,0 +1,54 @@
...
+    runtime="5 minute"
+    endtime=$(date -ud "$runtime" +%s)
+    FOUND=0
+
+    while [ ! "${FOUND}" -eq 1 ]; do
+        if [ "$(date -u +%s)" -ge "$endtime" ]; then
+            break
+            return 1
+        elif (journalctl --no-pager -u aktualizr-lite | grep "Pruning unused docker containers"); then
+            echo "Prune Successful"
```
mwasilew, line 19: I'd like a more detailed check here. Please see which containers/layers are present before and after the prune.

automated/linux/docker-prune/z-50-fioctl.toml
```diff
@@ -0,0 +1,13 @@
+{
```
mwasilew, line 1: this file needs to be renamed.

automated/linux/docker-prune/disable_prune.sh
```diff
@@ -0,0 +1,84 @@
...
+	echo "Sleeping 1s"
+	sleep 1
+	cat /var/sota/ota.signal
+	SIGNAL=$(</var/sota/ota.signal)
+	echo "SIGNAL: ${SIGNAL}."
+done
+report_pass "${TYPE}-update-post-received"
+
+if check_image_prune; then
+    echo "Image Prune successful"
```
roxell, line 69: Can this echo be dropped?

automated/linux/docker-prune/disable_prune.sh
```diff
@@ -0,0 +1,84 @@
...
+    report_pass "disable-prune"
+else
+    echo "Image Prune unsuccessful"
```
roxell, line 72: Can this echo be dropped?

automated/linux/docker-prune/disable_prune.sh
```diff
@@ -0,0 +1,84 @@
...
+do
+	echo "Sleeping 1s"
+	sleep 1
+	cat /var/sota/ota.signal
+	SIGNAL=$(</var/sota/ota.signal)
+	echo "SIGNAL: ${SIGNAL}."
+done
+report_pass "${TYPE}-update-post-received"
+
+if check_image_prune; then
```
roxell, line 68: what about pushing this into check into a funcition in prune-lib.sh ?

automated/linux/docker-prune/disable_prune.sh
```diff
@@ -0,0 +1,84 @@
+#!/bin/bash
```
roxell, line 1: maybe change this file to disable-prune.sh same with the yaml file?
There are more files with "-" instead of "_"....

automated/linux/docker-prune/disable_prune.sh
```diff
@@ -0,0 +1,84 @@
...
+else
+    systemctl enable --now lmp-device-auto-register || error_fatal "Unable to register device"
+fi
+
+while ! systemctl is-active aktualizr-lite; do
+    echo "Waiting for aktualizr-lite to start"
+    sleep 1
+done
+# add some delay so aklite can setup variables
+sleep 5
```
roxell, line 43: can all from line 20 to here be a function in prune-lib.sh too?

automated/linux/docker-prune/disable-prune.sh
```diff
@@ -0,0 +1,49 @@
...
+wait_for_signal
+
+if check_image_prune; then
+    report_pass "disable-prune"
+else
+    report_fail "disable-prune"
+fi
+
+if check_image shellhttpd; then
+    echo "Image present"
```
roxell, line 42: is this needed?
report_(fail|pass) do echo as well...

automated/linux/docker-prune/prune.sh
```diff
@@ -0,0 +1,43 @@
...
+
+OUTPUT="$(pwd)/output"
+RESULT_FILE="${OUTPUT}/result.txt"
+export RESULT_FILE
+
+! check_root && error_msg "You need to be root to run this script."
+create_out_dir "${OUTPUT}"
+
+if check_image hub.foundries.io/lmp-ci-testing-apps/shellhttpd; then
+    echo "image found"
```
roxell, line 14: same here...

automated/linux/docker-prune/prune.sh
```diff
@@ -0,0 +1,43 @@
...
+# shellcheck disable=SC1091
+. ../../lib/sh-test-lib
+. ./prune-lib.sh
+
+OUTPUT="$(pwd)/output"
+RESULT_FILE="${OUTPUT}/result.txt"
+export RESULT_FILE
+
+! check_root && error_msg "You need to be root to run this script."
+create_out_dir "${OUTPUT}"
```
roxell, line 11: maybe there are other bits in the files that can be placed in the lib too... I'm just asking.

automated/linux/docker-prune/prune-lib.sh
```diff
@@ -0,0 +1,92 @@
...
+            return 0
+        fi
+        sleep 1
+   done
+}
+
+compare_sha () {
+    if diff "$(pwd)/sha.txt" <(ls /var/sota/reset-apps/apps/shellhttpd) > /dev/null; then
+        return 1
+    else
```
roxell, line 50: assuming else isn't needed here, just return.

automated/linux/docker-prune/prune-lib.sh
```diff
@@ -0,0 +1,92 @@
...
+    fi
+}
+
+check_image () {
+    image_name="$1"
+
+    if [[ "$(docker images -q "${image_name}" 2> /dev/null)" == "" ]] ; then
+        echo "Docker image '${image_name}' not found."
+        return 1
+    else
```
roxell, line 21: else isn't needed just return.

automated/linux/docker-prune/prune-lib.sh
```diff
@@ -0,0 +1,92 @@
+#!/bin/bash
+
+verify_app_present () {
+    app="$1"
+
+    if docker ps | grep -q "${app}"; then
+        echo "${app} container found"
+        return 0
+    else
```
roxell, line 9: else isn't needed just return.

automated/linux/docker-prune/prune-lib.sh
```diff
@@ -0,0 +1,92 @@
...
+        return 0
+    fi
+}
+
+get_image_sha () {
+    ls /var/sota/reset-apps/apps/shellhttpd > "$(pwd)/sha.txt"
+}
+
+check_image_prune () {
+    runtime="5 minutes"
```
roxell, line 32: Does it make sense to have this as a variable that we can change from the yaml file?

brenwoo: Well, it's only really there as a maximum time to run before breaking the loop. In theory, it should never take that long unless there is a problem.

automated/linux/docker-prune/prune-lib.sh
brenwoo: Cool, I did not think to try this

automated/linux/docker-prune/disable-prune.sh
brenwoo: No, not really. I am just used to looking for my own echo messages rather than report_pass|fail. I will remove them

automated/linux/docker-prune/prune.sh
brenwoo: I can certainly have another look to move over anything else

automated/linux/docker-prune/prune-lib.sh
roxell: OK

PR #430, opened 2023-02-24
kvm-unit-tests: Fix naming of run_tests.sh

broonie:
The script is checking for run-tests.sh when looking to see if we have a
preinstalled copy of the tests but the script that the tests install is
run_tests.sh with an _ not a -.  Fix that.


nareshkamboju:
@broonie Thanks for submitting this patch.

PR #431, opened 2023-02-27
kselftest: Run the ftrace tests

broonie:
The ftrace tests are currently skipped with references to some kernel
bugs from back in 2017.  Given that developers use these tests regularly
without seeing problems and the age of the bugs it seems likely that
they are no longer relevant, testing seems to indicate that's the case
for at least arm64.  Remove the skip.


broonie:
Sample non-hanging job.

nareshkamboju (approved):

PR #432, opened 2023-03-07
automated: linux: add se05x reset after running xtest

mwasilew:
When running xtest on the device with secure element (SE050) some tests generate objects in the element. These objects are not deleted by the tests:
https://github.com/OP-TEE/optee_test/issues/652
In the long run secure element runs out of space and all tests start failing with strange errors. To mitigate this the script will call reset method on SE050 after all tests are completed. This should at least fix the problem of residual objects.

### DIFF
--- a/automated/linux/optee/optee-xtest.sh
+++ b/automated/linux/optee/optee-xtest.sh
@@ -5,16 +5,18 @@ OUTPUT="$(pwd)/output"
 RESULT_FILE="${OUTPUT}/result.txt"
 TEST_LEVEL="0"
 TEST_SUITE="regression"
+SE05X_TOOL=""
 
 usage() {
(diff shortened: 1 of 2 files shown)


automated/linux/optee/optee-xtest.sh
```diff
@@ -5,16 +5,18 @@ OUTPUT="$(pwd)/output"
 RESULT_FILE="${OUTPUT}/result.txt"
 TEST_LEVEL="0"
 TEST_SUITE="regression"
+SE05X_TOOL=""
 
 usage() {
     echo "Usage: $0 [-l <0-15> -t <regression|benchmark>]" 1>&2
```
roxell, line 11: Don't you want to update the usage too?

mwasilew:
@roxell I updated the PR. Could you take a look?

PR #434, opened 2023-03-10
utils: fix bad/illegal format or missing URL warning from curl

chase-qi:
CURL_VERBOSE_FLAG="" by default. `curl "" ...` works but reports `bad/illegal format or missing URL` warning. Removing the double quotes on the variable fixes the issue.


chase-qi:
@roxell @fboudra would you please take a look?

fboudra:
Thanks @roxell  :) you've been faster. TODO--

PR #438, opened 2023-04-03
noninteractive-tradefed: support to use lkft-cache for cache

liuyq:
only lkft-cache.lkftlab is supported with this PR, it should be more generic when another kisscache server is supported.

### DIFF
--- a/automated/android/noninteractive-tradefed/tradefed.sh
+++ b/automated/android/noninteractive-tradefed/tradefed.sh
@@ -62,12 +62,12 @@ disable_suspend
 # wait_homescreen "${TIMEOUT}"
 
 # Download CTS/VTS test package or copy it from local disk.
-if echo "${TEST_URL}" | grep "^http" ; then
-    wget -S --progress=dot:giga "${TEST_URL}"
-else
-    cp "${TEST_URL}" ./
-fi
 file_name=$(basename "${TEST_URL}")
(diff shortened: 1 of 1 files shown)


automated/android/noninteractive-tradefed/tradefed.sh
```diff
@@ -62,12 +62,12 @@ disable_suspend
 # wait_homescreen "${TIMEOUT}"
 
 # Download CTS/VTS test package or copy it from local disk.
+file_name=$(basename "${TEST_URL}")
 if echo "${TEST_URL}" | grep "^http" ; then
-    wget -S --progress=dot:giga "${TEST_URL}"
+    wget -S --progress=dot:giga "${TEST_URL}" -O "${file_name}"
 else
-    cp "${TEST_URL}" ./
+    cp "${TEST_URL}" "./${file_name}"
```
roxell, line 69: Maybe you should move the file and not do a copy of it?

liuyq: > Maybe you should move the file and not do a copy of it?

first, this change is only to make sure that the file is copied/downloaded to a file with the specified name.
second, the cp branch here is for the case that run locally, if mv is used, the file might not be found again when run next time.

roxell: ok

PR #444, opened 2023-05-31
optee-xtest.yml: clear storage before executing tests

mikkorapeli-linaro:
Previously failed tests can leave some data in storage which fails the tests later on unless data is cleared. --clear-storage option is new to xtest:

https://github.com/OP-TEE/optee_test/pull/670

See also:

https://gitlab.com/Linaro/trustedsubstrate/meta-ledge-secure/-/merge_requests/60

Ignore return values since the option may not exist in all xtest versions and it can also report failures.


jforissier:
LGTM

PR #445, opened 2023-06-12
automated: linux: avoid OTA in factory-reset test

mwasilew:
When performing factory reset it is assumed the device is up to date with respect to aktualizr-lite and ostree. This patch allows to add a tag to the sota.toml which forces aktualizr-lite to use this tag. This gives a chance to add a tag to only one target on the backend side and avoid OTA update during the test.


mwasilew:
Example job: https://lava.infra.foundries.io/scheduler/job/23059

PR #449, opened 2023-06-30
automated: linux: add xenomai4 test suite

mwasilew:
This patch adds a smoke test for xenomai4 RT solution for kernel. By default the test runs a list of tests predefined in 'evl' tool. The output is PASS/FAIL/SKIP for each test. Tests report SKIP when the required configs are not available in the kernel.


mwasilew:
Example test job: https://lava.infra.foundries.io/scheduler/job/24519#L3109

PR #452, opened 2023-07-12
automated: linux: factory-reset: fix device registration

mwasilew:
Enable proper auto-registration with tags. Note: this patch requires support in the lmp-device-auto-register script.


roxell:
Thanks for updating with the Link to the other PR in the changelog.
Please reword the changelog again as we spoke about on IRC, since it only supports a tag not multiple.

mwasilew:
Fixed and now compatible with LmP patch

PR #453, opened 2023-07-13
kselftest: Adding new test case for sharding status

nareshkamboju:
Adding new test case to check sharding status.
When sharding file is empty, test will fail and exit.

Suggested-by:  Anders Roxell <anders.roxell@linaro.org>

### DIFF
--- a/automated/linux/kselftest/kselftest.sh
+++ b/automated/linux/kselftest/kselftest.sh
@@ -200,6 +200,12 @@ elif [ -n "${TST_CMDFILES}" ]; then
         echo "============== Tests to run ==============="
         cat shardfile
         echo "===========End Tests to run ==============="
+        if [ -s shardfile ]; then
+            report_pass "shardfile-${test}"
+        else
+            report_fail "shardfile-${test}"
+            continue
+        fi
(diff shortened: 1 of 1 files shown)


automated/linux/kselftest/kselftest.sh
```diff
@@ -200,6 +200,12 @@ elif [ -n "${TST_CMDFILES}" ]; then
         echo "============== Tests to run ==============="
         cat shardfile
         echo "===========End Tests to run ==============="
+        if [ -s shardfile ]; then
+            report_pass "shardfile-${TST_CMDFILES}"
```
roxell, line 204: this should be ${test} and not ${TST_CMDFILES} or?

nareshkamboju: ${TST_CMDFILES} is arm64
${test} is arm64:fp-stress

We are trying to detect the when sharding file size is zero that mean the total subtests get missed.

nareshkamboju: @roxell  Thanks for the review comments.
Now those review comments addressed and pushed new version of patch.

PR #454, opened 2023-07-26
noninteractive-tradefed: change to work with python3

liuyq:
This commit change to use python3 for tradefed-runner.py,
and drop the packages installation as they will be installed
on the Ubuntu20.04 based docker image side which will have
python3 installed as the default python.

The background is that, the latest AOSP vts package needs to
have python3.8+ to run the vts_kernel_proc_file_api_test module
after the following commit:
    https://android.googlesource.com/platform/test/vts-testcase/vndk/+/6f09ccb69d553a48dd231635a06889af26ff545f%5E%21/#F3

Otherwise, the following error will be reported:
```
    07-26 12:45:26 I/ModuleListener: [66/67] cbe37e vts_kernel_proc_file_api_test __main__.VtsKernelProcFileApiTest#test_ProcZoneInfoTest FAILURE: Traceback (most recent call last):
      File "/tmp/Soong.python_p2c4wyvb/vts_kernel_proc_file_api_test.py", line 277, in test
        test_object.set_api_level(self.dut)
      File "/tmp/Soong.python_p2c4wyvb/proc_tests/KernelProcFileTestBase.py", line 140, in set_api_level
        self.api_level = dut.getLaunchApiLevel(strict=False)
      File "/tmp/Soong.python_p2c4wyvb/vts/testcases/vndk/utils.py", line 147, in getLaunchApiLevel
        return self.GetLaunchApiLevel()
      File "/tmp/Soong.python_p2c4wyvb/vts/testcases/vndk/utils.py", line 122, in GetLaunchApiLevel
        level_str = self._GetProp("ro.product.first_api_level")
      File "/tmp/Soong.python_p2c4wyvb/vts/testcases/vndk/utils.py", line 94, in _GetProp
        out, err, return_code = self.Execute("getprop", name)
      File "/tmp/Soong.python_p2c4wyvb/vts/testcases/vndk/utils.py", line 68, in Execute
        cmd = ["adb", "-s", self._serial_number, "shell", shlex.join(args)]
    AttributeError: module 'shlex' has no attribute 'join'
```

### DIFF
--- a/automated/android/noninteractive-tradefed/setup.sh
+++ b/automated/android/noninteractive-tradefed/setup.sh
@@ -5,22 +5,15 @@
 . ../../lib/sh-test-lib
 . ../../lib/android-test-lib
 
-JDK="openjdk-11-jdk-headless"
 java_path="/usr/lib/jvm/java-11-openjdk-amd64/bin/java"
 if [ -n "${ANDROID_VERSION}" ] && echo "${ANDROID_VERSION}" | grep -q  "aosp-android14"; then
     # use openjdk-17 for Android14+ versions
-    JDK="openjdk-17-jdk-headless"
     java_path="/usr/lib/jvm/java-17-openjdk-amd64/bin/java"
 fi
-
-PKG_DEPS="coreutils usbutils curl wget zip xz-utils python-lxml python-setuptools python-pexpect aapt lib32z1-dev libc6-dev-i386 lib32gcc1 libc6:i386 libstdc++6:i386 libgcc1:i386 zlib1g:i386 libncurses5:i386 python-dev python-protobuf protobuf-compiler python-virtualenv python-pip python-pexpect psmisc"
 
 dist_name
 case "${dist}" in
     ubuntu)
-        dpkg --add-architecture i386
-        apt-get update -q
-        install_deps "${PKG_DEPS} ${JDK}"
         # make sure to use the right java version
         update-alternatives --set java ${java_path}
         ;;
--- a/automated/android/noninteractive-tradefed/tradefed-runner.py
+++ b/automated/android/noninteractive-tradefed/tradefed-runner.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 import datetime
 import os
@@ -27,7 +27,7 @@ ATOMIC = "atomic"
 
 def result_parser(xml_file, result_format):
     etree_file = open(xml_file, "rb")
-    etree_content = etree_file.read()
+    etree_content = etree_file.read().decode("utf-8")
     rx = re.compile("&#([0-9]+);|&#x([0-9a-fA-F]+);")
     endpos = len(etree_content)
     pos = 0
--- a/automated/android/noninteractive-tradefed/tradefed.yaml
+++ b/automated/android/noninteractive-tradefed/tradefed.yaml
@@ -14,7 +14,6 @@ metadata:
         - functional
 
 params:
-    SKIP_INSTALL: "false"
     # Specify timeout in seconds for wait_boot_completed and wait_homescreen.
     TIMEOUT: "300"
     # Download CTS package or copy it from local disk.


bhcopeland:
Surely it makes sense to run a distro that comes with Python 3.8 by default (i.e. Ubuntu 20.04 or latest Debian) than supporting a older distro to use a new python version.

liuyq:
> Surely it makes sense to run a distro that comes with Python 3.8 by default (i.e. Ubuntu 20.04 or latest Debian) than supporting a older distro to use a new python version.

It's the next thing I am working on. https://review.linaro.org/c/ci/dockerfiles/+/45051, but there are errors I need to fix one by one, and need to test it for different Android versions as well, which might need some time.
If anyone could help submit a change to create the Ubuntu 20.04 based docker image quickly, it would be appreciated.

bhcopeland:
Thanks Yongqin. I would say lets get the Dockerfile approach through and drop this review. I have left you a comment on the dockerfile and we can get that through. More than happy to help with the PR there.

liuyq:
> Thanks Yongqin. I would say lets get the Dockerfile approach through and drop this review. I have left you a comment on the dockerfile and we can get that through. More than happy to help with the PR there.

Thanks for checking the change for the dockerfiles repository and this one, while the dockerfiles change is being worked on, I still like to have this workaround in first, which would help us have enough time to make that one perfect.

liuyq:
> > Thanks Yongqin. I would say lets get the Dockerfile approach through and drop this review. I have left you a comment on the dockerfile and we can get that through. More than happy to help with the PR there.

Another reason I'd like to have this in, is not to bind this script to some specific docker image, whether linaro docker image or not, even in a real dist instead of a docker container.

bhcopeland:
Hi Yongqin,

I have pushed your docker image through that now supports the correct version of Python. 

As for this changeset, it can be dropped now. I see no reason to push this through. Moreover most the deps in setup.sh should be pushed back to the base image, and changing python versions at this level doesn't make sense to me.

liuyq:
> As for this changeset, it can be dropped now. I see no reason to push this through. Moreover most the deps in setup.sh should be pushed back to the base image, and changing python versions at this level doesn't make sense to me.

As I said in another comment, this way the test script will be bound to the docker image, separate installations will be required when like running on the host without using the docker. Do other test scripts have this requirements?

automated/android/noninteractive-tradefed/setup.sh
```diff
@@ -13,14 +13,12 @@ if [ -n "${ANDROID_VERSION}" ] && echo "${ANDROID_VERSION}" | grep -q  "aosp-and
     java_path="/usr/lib/jvm/java-17-openjdk-amd64/bin/java"
 fi
 
-PKG_DEPS="coreutils usbutils curl wget zip xz-utils python-lxml python-setuptools python-pexpect aapt lib32z1-dev libc6-dev-i386 lib32gcc1 libc6:i386 libstdc++6:i386 libgcc1:i386 zlib1g:i386 libncurses5:i386 python-dev python-protobuf protobuf-compiler python-virtualenv python-pip python-pexpect psmisc"
-
```
roxell, line 17: why removing all the PKG_DEPS?

automated/android/noninteractive-tradefed/setup.sh
```diff
@@ -13,14 +13,12 @@ if [ -n "${ANDROID_VERSION}" ] && echo "${ANDROID_VERSION}" | grep -q  "aosp-and
...
 dist_name
 case "${dist}" in
     ubuntu)
         dpkg --add-architecture i386
         apt-get update -q
-        install_deps "${PKG_DEPS} ${JDK}"
+        install_deps "${JDK} ${PYTHON38}"
```
roxell, line 21: Where do the variable PYTHON38 come from?

roxell:
instead of this, why not just check if VERSION_ID less than 20.04 error out and print a relevant message.

automated/android/noninteractive-tradefed/setup.sh
liuyq: As you can see from @bhcopeland 's sugggestion, the packages are all installed within the docker image, no need to install them here again.

automated/android/noninteractive-tradefed/setup.sh
liuyq: it's removed now.

automated/android/noninteractive-tradefed/setup.sh
roxell: right, but that make sense for you. but if I want to run this and I don't have it preinstalled I still need the PKG_DEPS.
So since you force people to use ubuntu, just check that its at least ubuntu:20.04... 
1. Then just check that its ubuntu:20.04.
2. default in your job submission SKIP_INSTALL: 'true'
3. if possible add the java-bits into the docker container ?
4. but that still needs to be here if its not installed...

liuyq: 1. it's better not restricted to 20.04, without the check, it would be possible to be run within docker image like based Ubuntu 22.04
2. dropped SKIP_INSTALL, as it's not used anymore
3. JDK will be installed in the docker images side.
4. it will make the script here complicated to support more host versions, so drop the all the packages installation here, and let the user to prepare the host environments.

liuyq:
tested with the folllowing jobs:
```
Android11:
    https://lkft.validation.linaro.org/scheduler/job/6636839    lkft-android-android11-5.4-ubuntu20.04-cts-lkft
        Tests Passed	63966
        Tests Failed	60
        Modules Done	28
        Modules Total	28
    https://lkft.validation.linaro.org/scheduler/job/6636840    lkft-android-android11-5.4-ubuntu20.04-vts-kernel-arm64-v8a
        Tests Passed	1378
        Tests Failed	41
        Modules Done	19
        Modules Total	19
    https://lkft.validation.linaro.org/scheduler/job/6636841    lkft-android-android11-5.4-ubuntu20.04-vts-kernel-armeabi-v7a
        Tests Passed	1439
        Tests Failed	43
        Modules Done	19
        Modules Total	19

Android14:
    https://lkft.validation.linaro.org/scheduler/job/6636842    lkft-android-android14-6.1-ubuntu20.04-cts-lkft
        Tests Passed	79846
        Tests Failed	4853
        Modules Done	25
        Modules Total	27
    https://lkft.validation.linaro.org/scheduler/job/6637186    lkft-android-android14-6.1-ubuntu20.04-vts-kernel-arm64-v8a
        Tests Passed	1550
        Tests Failed	1
        Modules Done	20
        Modules Total	20
    https://lkft.validation.linaro.org/scheduler/job/6636844    lkft-android-android14-6.1-ubuntu20.04-vts-kernel-armeabi-v7a
        Tests Passed	1507
        Tests Failed	1
        Modules Done	20
        Modules Total	20

Main:
    https://lkft.validation.linaro.org/scheduler/job/6636845    lkft-android-android-mainline-ubuntu20.04-cts-lkft
        Tests Passed	83139
        Tests Failed	91
        Modules Done	27
        Modules Total	27
    https://lkft.validation.linaro.org/scheduler/job/6636846    lkft-android-android-mainline-ubuntu20.04-vts-kernel-arm64-v8a
        Tests Passed	1548
        Tests Failed	1
        Modules Done	20
        Modules Total	20
    https://lkft.validation.linaro.org/scheduler/job/6636847    lkft-android-android-mainline-ubuntu20.04-vts-kernel-armeabi-v7a
        Tests Passed	1513
        Tests Failed	1
        Modules Done	20
        Modules Total	20
```

liuyq:
Hi, @roxell @bhcopeland 
Please feel free to continue commenting and submitting  improvements/fixes for this test.
Let me merge this change first for now.

PR #455, opened 2023-08-01
automated: linux: rcutorture: drop config checking

roxell:
No need to check for fragments since it can be built into the kernel and then run on boot.


chase-qi:
The new changes looks to me too.

nareshkamboju (approved):

PR #465, opened 2023-10-02
automated: linux: Add AWS IoT use case to el2go test

mwasilew:
This patch adds support for testing AWS IoT integration for EL2GO service with FoundriesFactory. More detailed instructions are available from the el2go.yaml


mwasilew:
It's set in /etc/os-release. This is the FoundriesFactory name for the build.

mwasilew:
I added a comment in the shell script to clarify the situation

nareshkamboju (changes requested):
Since check returns is checking.
then no problem.

+ #echo "Cleanup SE050"
you may remove this line.
or
keep it.
 echo "Cleanup SE050"

mwasilew:
Fixed.

PR #466, opened 2023-10-05
automated: linux: ltp: skipfile-lkft: Remove the all_boards alias

katieworton:
Previously the `all` keyword for boards was replaced with an list `all_boards` to store the list of supported boards to make it possible to remove boards as part of the automation for skipfile testing.

To make the commits easier to read, it has been suggested that the all_boards alias should be removed, and the list of boards should be maintained within each skipfile entry. This way when boards are removed as part of the automated skipfile testing the diff will be much easier to read. When a board is removed from an entry that previously had all boards, instead of the diff reading "all_boards" -> "\<list of all_boards with some entries removed\>" the diff will just show which boards were removed.

This will make the update commits much easier to interpret and review. Additionally, as the automated skipfile testing is run and boards are removed based on the test results, the all_boards variable will likely be removed in most places in the skipfile, as most tests will likely work on at least one board. For these reasons, it seems sensible to remove the alias and variable.

### DIFF
--- a/automated/linux/ltp/skipfile-lkft.yaml
+++ b/automated/linux/ltp/skipfile-lkft.yaml
@@ -19,39 +19,32 @@
 #   List of tests to skip
 
 
-globals:
-# This list represents the devices in the lab from lkft.validation.linaro.org,
-# and the qemu devices available in tuxsuite.
-  - boards: &all_boards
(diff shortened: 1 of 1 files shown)


nareshkamboju (approved):
Thanks for the PR.
Good work.

PR #467, opened 2023-10-09
noninteractive-tradefed: print more network debug information

liuyq:
as discussed with the lab team


roxell:
is this really needed here?
Can't we test and debug this somewhere else?

liuyq:
thanks for checking it, @roxell !
I will test it with my fork repository.

PR #468, opened 2023-10-10
automated: linux: fix network-basic exit code

mwasilew:
When last test in network-basic fails (download-a-file) the whole set of tests is lost because LAVA receives exit code 1 and ignores the whole network-basic suite. This patch adds "exit 0" to the end of the test script to avoid this issue.


roxell:
sorry for the delay, lgtm.

PR #480, opened 2024-01-18
automated: linux: fix issues with static-delta OTA update

mwasilew:
When updating the device using static deltas, the test script has to wait for aktualizr-lite to complete installation before rebooting.

When there is no static detla available, the test job should end as incomplete.

This patch also changes how the aklite signal is checked to align with other OTA testing scripts.


roxell:
Can we not use tabs when indenting, this file is indented with space.

mwasilew:
Hmm, I didn't change the indentation. I checked a few other scripts and all are using spaces. Are you sure you want to change it?

roxell:
> Hmm, I didn't change the indentation. I checked a few other scripts and all are using spaces. Are you sure you want to change it?

hmm, I'm sorry my comment didn't make sense. What I meant was that you have added tabs in this PR, and the script uses spaces.

mwasilew:
Now it makes sense. Fixed.

PR #482, opened 2024-01-19
automated: linux: device registration with HSM module

mwasilew:
This patch allows to register device to FoundriesFactory using HSM. It's enabled in ota-update and factory-reset tests. Main use case is to test registration with SE050 devices from NXP.

### DIFF
--- a/automated/linux/factory-reset/prepare-reset.sh
+++ b/automated/linux/factory-reset/prepare-reset.sh
@@ -11,11 +11,14 @@ TYPE="factory_reset"
 ADDITIONAL_TYPE=""
 LABEL=""
 SOTA_CONFDIR="/etc/sota/conf.d"
+HSM_MODULE=""
 
 usage() {
     echo "\
     Usage: $0 [-t <factory_reset|factory_reset_keep_sota|factory_reset_keep_sota_docker>]
               [-a <factory_reset|factory_reset_keep_sota|factory_reset_keep_sota_docker>]
+              [-l <target label>]
+              [-S <hsm module>]
 
     -t <factory_reset|factory_reset_keep_sota|factory_reset_keep_sota_docker>
         factory_reset: Full reset, removes contents of /etc/ and /var/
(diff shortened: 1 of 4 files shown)


automated/linux/factory-reset/prepare-reset.sh
```diff
@@ -11,11 +11,14 @@ TYPE="factory_reset"
...
 LABEL=""
 SOTA_CONFDIR="/etc/sota/conf.d"
+HSM_MODULE=""
 
 usage() {
     echo "\
     Usage: $0 [-t <factory_reset|factory_reset_keep_sota|factory_reset_keep_sota_docker>]
               [-a <factory_reset|factory_reset_keep_sota|factory_reset_keep_sota_docker>]
+              [-l <target label>]
+              [-s <hsm module>]
```
roxell, line 21: Would it make sense to set this flag as '-S', since its that in automated/linux/ota-update/download-update.sh ?

mwasilew: Good idea. Done

roxell: Think you missed to change it from '[-s <hsm module>]' to '[-S <hsm module>]' here.

automated/linux/ota-update/download-update.sh
```diff
@@ -17,10 +17,11 @@ U_BOOT_VARIABLE_NAME="foobar"
 U_BOOT_VARIABLE_VALUE="baz"
 DEBUG="false"
 SOTA_CONFDIR="/etc/sota/conf.d"
+HSM_MODULE=""
 
 usage() {
-	echo "\
-	Usage: $0 [-t <kernel|uboot>] [-u <u-boot var read>] [-s <u-boot var set>] [-o <ostree|ostree+compose_apps>] [-V <variable name>] [-w <variable value>] [-d <true|false> ]
+    echo "\
+    Usage: $0 [-t <kernel|uboot>] [-u <u-boot var read>] [-s <u-boot var set>] [-o <ostree|ostree+compose_apps>] [-V <variable name>] [-w <variable value>] [-d <true|false> ]
```
roxell, line 24: Should '[-S <hsm module>]' be added here too ?

PR #487, opened 2024-01-25
docs: fix automated docs generation

mwasilew:
New version of material theme and plugins breaks when there is empty "tags" stanza in the metadata section of the markdown. This patch only adds "tags" if there are any to add.

Second issue fixed is adding an empty tags.md file which is now required to present collected tags in the generated docs.


roxell:
LGTM, thanks for fixing this.

PR #488, opened 2024-01-25
docs: bump version of mkdocs plugin

mwasilew:


mwasilew:
Plugin already uploaded to pypi.

PR #489, opened 2024-02-05
mkdocs_plugin: sort resulting docs alphabetically

mwasilew:
When docs are built in readthedocs the resulting test list is not sorted. This patch ensures the resulted rendered docs present the list in alphabetical order.


mwasilew:
Plugin needs to be updated in pypi when this patch is accepted.

mwasilew:
Uploaded version 1.4 to pypi. From the next merge docs should look better.

PR #493, opened 2024-02-10
automated: kselftest: Rewrite parser to support nesting

ryanhrob:
The previous kselftests TAP parser has a few shortcomings that prevent it from working well with the mm kselftests output. mm kselftests are nested up to 3 levels deep, with the most important information, from the user's perspective, usually contained in the middle level. But the parser isn't nesting-aware, and instead flattens test results to include the first and last level of nesting and ignroed everything in the middle. This leads to undescriptive and confusing test names in kernelci UI.

Additionally, conflicting test names extracted by the parser are not made unique, which leads to multiple distinct tests being collapsed into a single test within the kernelci UI, and its status is set to the last occurence of the test in the list. So you might have multiple instances that fail, but if the last one passes, it is shown as a single test that passes. This problem is compounded by the parser's inability to properly nest because the important middle level information is lost and this makes many more test names look identical, so even more get collapsed into one.

Solve all of this by rewriting the parser to properly support recursive parsing. The tree of tests are then flattened into a test list in depth-first order, where the test name is built from the name of each level. Further, if duplicate test names exist, append a "_dup<N>" to the second instance onwards, where N is a unique number. This guarrantees that all test points output by TAP appear in kernelci UI.

I've tested this against the output for arm64, ftrace, kvm, and sigstack kselftests (which don't have much nesting so work fine with the old parser): The outputs from both parsers are identical, except in a couple of instances where there are duplicate test name outputs and the new parser appends the "_dup<N>" suffix to make it unique.

I've also tested this against the output from the mm kselftests: The output from the new parser is as expected, and much more useful than the old parser.

The downside is that this implementation depends on the tap.py module (https://tappy.readthedocs.io). It is packaged for Debian and Ubuntu, so I've added that package as a dependency. But I couldn't find anything for Centos or Fedora, so this module (and its dependencies) will likely need to be installed from PyPI in these environments:

  $ pip3 install tap.py more-itertools pyyaml

### DIFF
--- a/automated/linux/kselftest/kselftest.sh
+++ b/automated/linux/kselftest/kselftest.sh
@@ -135,7 +135,7 @@ install() {
     dist_name
     # shellcheck disable=SC2154
     case "${dist}" in
-        debian|ubuntu) install_deps "sed perl wget xz-utils iproute2" "${SKIP_INSTALL}" ;;
+        debian|ubuntu) install_deps "sed perl wget xz-utils iproute2 python3-tap" "${SKIP_INSTALL}" ;;
         centos|fedora) install_deps "sed perl wget xz iproute" "${SKIP_INSTALL}" ;;
         unknown) warn_msg "Unsupported distro: package install skipped" ;;
     esac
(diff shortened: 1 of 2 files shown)


ryanhrob:
I've fixed this up to satisfy the linter. Is it possible to rerun the workflow?

roxell (approved):
Thank you @ryanhrob for the patch.

PR #495, opened 2024-02-15
Actions: Remove python-dep

bhcopeland:
We shouldn't lock down to old python version.


mwasilew:
IIRC there was a reason for this version (don't remember the details). However it looks the checks pass without the legacy python :)

mwasilew (approved):

roxell (approved):

PR #496, opened 2024-02-19
noninteractive-tradefed: stop using the latest adb

liuyq:
instead using the default version in the system,
either installed by the user manually or installed
inside some docker image, to avoid confusions caused by
unmanaged adb versions, like the "ADB server didn't ACK"
issue with the current latest adb version(platform-tools r35)
reported here:
https://lkft.validation.linaro.org/scheduler/job/7266683#L15671

### DIFF
--- a/automated/android/noninteractive-tradefed/setup.sh
+++ b/automated/android/noninteractive-tradefed/setup.sh
@@ -22,7 +22,6 @@ case "${dist}" in
         ;;
 esac
 
-install_latest_adb
 initialize_adb
 adb_root
 


liuyq:
HI, @bhcopeland @roxell 
This is also for the "ADB server didn't ACK" problem reported by jobs like here: https://lkft.validation.linaro.org/scheduler/job/7266683#L15671
Please help check and merge it in case you don't have any other concerns.

liuyq:
share with @sumitsemwal as well

roxell:
Can you please elaborate how this fixes the issue, and what is the issue in your commit message.

liuyq:
> Can you please elaborate how this fixes the issue, and what is the issue in your commit message.

updated the commit message now, the "ADB server didn't ACK" issue is related to the latest adb version(platform-tools r35),
so we need to stop using the latest one and change to use one known working version.
Please help review it again.

liuyq:
Hi, @bhcopeland @roxell 

Just want to check, do you have any other concerns on this change to have it merged?

roxell:
Won't it be better to pin the version rather than relay on the version on the dispatcher?

bhcopeland:
Yongqin it would have probably helped to link to the docker image here for context https://review.linaro.org/c/ci/dockerfiles/+/46721

roxell (approved):

roxell:
oh thank you @bhcopeland I missed that dockerfile change.

Issue #500, opened 2024-03-05
kselftest: add support for hosting tarball in Azure Files


pawiecz:
URLs to files stored in Azure Files [include SAS token](https://learn.microsoft.com/en-us/rest/api/storageservices/create-account-sas) ([example](https://kciapistagingstorage1.file.core.windows.net/staging/kbuild-gcc-10-arm64-65e5cb927bef584b36e0f89f/kselftest.tar.gz?sv=2022-11-02&ss=f&srt=sco&sp=r&se=2024-10-17T19:19:12Z&st=2023-10-17T11:19:12Z&spr=https&sig=sLmFlvZHXRrZsSGubsDUIvTiv%2BtzgDq6vALfkrtWnv8%3D)).

[Extracting such an archive](https://github.com/Linaro/test-definitions/blob/a14c1633b22acb33ae3149c88ea644e7a724424b/automated/linux/kselftest/kselftest.sh#L158) currently fails ([example LAVA TestJob](https://lava.collabora.dev/scheduler/job/12933906#L1235)) because name of the downloaded tarball includes SAS token.

Possible solution: rename [`wget` output](https://github.com/Linaro/test-definitions/blob/a14c1633b22acb33ae3149c88ea644e7a724424b/automated/linux/kselftest/kselftest.sh#L157) (PR will follow-up shortly).

pawiecz:
Discarded idea: drop SAS using `basename` option for suffix removal (`-s`) - won't work as it doesn't support wildcards and the token may change

PR #501, opened 2024-03-06
automated: kselftest: explicitly set name of downloaded tarball

pawiecz:
This patch always renames downloaded kselftest tarball to the value set in "TESTPROG". Default behaviour of the kselftest.sh script remains unchanged:

1) TESTPROG and TESTPROG_URL undefined: fall back to tarball/URL defaults
2) only TESTPROG defined: download specified test data from Linaro
3) only TESTPROG_URL defined: rename downloaded tarball to one of the defaults (based on architecture) but extract it immediately (so it's discarded anyway and the name becomes irrelevant)
4) both defined: specify name for the downloaded tarball

Case (4) is the only new branch added by this patch. Adjustments in case (3) should be transparent when running kselftest.sh script.

Fixes: #500

### DIFF
--- a/automated/linux/kselftest/kselftest.sh
+++ b/automated/linux/kselftest/kselftest.sh
@@ -154,8 +154,8 @@ if [ -d "${KSELFTEST_PATH}" ]; then
 else
     # Fetch whatever we have been aimed at, assuming only that it can
     # be handled by "tar". Do not assume anything about the compression.
-    wget "${TESTPROG_URL}"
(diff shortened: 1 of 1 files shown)


pawiecz:
This patch has been verified in Collabora's LAVA lab: https://lava.collabora.dev/scheduler/job/12974384#L1205: tarball downloaded from Azure Files was renamed and successfully extracted.

(linked LAVA TestJob logs will be pruned from database in 120 days due to data retention policy)

PR #511, opened 2024-06-05
kselftest: Add new option -E for specifying environment variables

crazoes:
Currently we are not able to run the subtests of a kselftest due to lack of support for passing the parameters to the test case. Hence, solve this problem by adding support for passing environment variables in KernelCI kselftest.

### DIFF
--- a/automated/linux/kselftest/kselftest.sh
+++ b/automated/linux/kselftest/kselftest.sh
@@ -44,11 +44,12 @@ usage() {
                     [-S kselftest-skipfile]
                     [-b board]
                     [-g branch]
-                    [-e environment]" 1>&2
+                    [-e environment]
+                    [-E ENV_NAME=ENV_VAL]" 1>&2
     exit 1
 }
 
-while getopts "i:n:c:T:t:s:u:p:L:S:b:g:e:h" opt; do
+while getopts "i:n:c:T:t:s:u:p:L:S:b:g:e:E:h" opt; do
     case "${opt}" in
         i) SHARD_INDEX="${OPTARG}" ;;
         n) SHARD_NUMBER="${OPTARG}" ;;
@@ -106,6 +107,13 @@ while getopts "i:n:c:T:t:s:u:p:L:S:b:g:e:h" opt; do
             ;;
         e)
             export ENVIRONMENT="${OPTARG}"
+            ;;
+        E)
+            if [ -n "${OPTARG}" ]; then
+                set -x
+                eval "export ${OPTARG}"
+                set +x
+            fi
             ;;
         h|*) usage ;;
     esac
--- a/automated/linux/kselftest/kselftest.yaml
+++ b/automated/linux/kselftest/kselftest.yaml
@@ -50,6 +50,9 @@ params:
     BOARD: ""
     BRANCH: ""
     ENVIRONMENT: ""
+    # Set arbitrary environment variables. The value set here is directly evaluated
+    # as `eval "export ${ENV}"`.
+    ENV: ""
 
     # Number of shards that will be done, default 1 which is the same as no sharding.
     SHARD_NUMBER: 1
@@ -60,5 +63,5 @@ params:
 run:
     steps:
         - cd ./automated/linux/kselftest/
-        - ./kselftest.sh -c "${TST_CMDFILES}" -T "${TST_CASENAME}" -t "${TESTPROG}" -s "${SKIP_INSTALL}" -u "${TESTPROG_URL}" -L "${SKIPLIST}" -S "${SKIPFILE}" -b "${BOARD}" -g "${BRANCH}" -e "${ENVIRONMENT}" -p "${KSELFTEST_PATH}" -n "${SHARD_NUMBER}" -i "${SHARD_INDEX}"
+        - ./kselftest.sh -c "${TST_CMDFILES}" -T "${TST_CASENAME}" -t "${TESTPROG}" -s "${SKIP_INSTALL}" -u "${TESTPROG_URL}" -L "${SKIPLIST}" -S "${SKIPFILE}" -b "${BOARD}" -g "${BRANCH}" -e "${ENVIRONMENT}" -p "${KSELFTEST_PATH}" -n "${SHARD_NUMBER}" -i "${SHARD_INDEX}" -E "${ENV}"
         - ../../utils/send-to-lava.sh ./output/result.txt


automated/linux/kselftest/kselftest.sh
```diff
@@ -107,6 +108,20 @@ while getopts "i:n:c:T:t:s:u:p:L:S:b:g:e:h" opt; do
         e)
             export ENVIRONMENT="${OPTARG}"
             ;;
+        E)
+            if [ -n "$OPTARG" ]; then
```
roxell, line 112: This shouldn't be needed.

automated/linux/kselftest/kselftest.sh
```diff
@@ -107,6 +108,20 @@ while getopts "i:n:c:T:t:s:u:p:L:S:b:g:e:h" opt; do
...
+                 IFS='=' read -r env_name env_val << EOF
+$OPTARG
+EOF
```
roxell, line 115: I think this would be enough.

IFS='=' read -r env_name env_val <<< "${OPTARG}"

automated/linux/kselftest/kselftest.yaml
```diff
@@ -50,6 +50,8 @@ params:
     BOARD: ""
     BRANCH: ""
     ENVIRONMENT: ""
+    # Set arbitrary environment variable
```
roxell, line 53: Is it "arbitrary environment variables" or is there an example you can give here too for clarity?

roxell (commented):
Thank you for the patch.

Can you please share an example how it can be used.

Also can you add your Signed-off-by.

automated/linux/kselftest/kselftest.sh
crazoes: yeah i had that initially, but shellcheck warns on that

automated/linux/kselftest/kselftest.sh
crazoes: This is to allow empty args, which will be the default for most tests with `-E ${ENV}` where `ENV` is not set. Without this the `-z` checks on the env_name and env_val will cause the script to exit.

automated/linux/kselftest/kselftest.yaml
crazoes: for the yaml it is single variable the way parsing is done and how the argument passing is hardcoded to `-E ${ENV}`, for the script user can pass `-E <env1> -E <env2>` multiple times for each env var

roxell: Would it be possible to add an example into the # comment?

automated/linux/kselftest/kselftest.sh
roxell: OK understood, thank you.

automated/linux/kselftest/kselftest.sh
```diff
@@ -107,6 +108,20 @@ while getopts "i:n:c:T:t:s:u:p:L:S:b:g:e:h" opt; do
...
+                if [ -z "$env_name" ] || [ -z "$env_val" ]; then
```
roxell, line 116: is there a reason for "$env_name" here and not "${env_name}" but doing that further down?

automated/linux/kselftest/kselftest.sh
roxell: Can "$OPTARG" be "${OPTARG}" like the others please.

automated/linux/kselftest/kselftest.yaml
```diff
@@ -60,5 +69,5 @@ params:
 run:
     steps:
         - cd ./automated/linux/kselftest/
-        - ./kselftest.sh -c "${TST_CMDFILES}" -T "${TST_CASENAME}" -t "${TESTPROG}" -s "${SKIP_INSTALL}" -u "${TESTPROG_URL}" -L "${SKIPLIST}" -S "${SKIPFILE}" -b "${BOARD}" -g "${BRANCH}" -e "${ENVIRONMENT}" -p "${KSELFTEST_PATH}" -n "${SHARD_NUMBER}" -i "${SHARD_INDEX}"
+        - ./kselftest.sh -c "${TST_CMDFILES}" -T "${TST_CASENAME}" -t "${TESTPROG}" -s "${SKIP_INSTALL}" -u "${TESTPROG_URL}" -L "${SKIPLIST}" -S "${SKIPFILE}" -b "${BOARD}" -g "${BRANCH}" -e "${ENVIRONMENT}" -p "${KSELFTEST_PATH}" -n "${SHARD_NUMBER}" -i "${SHARD_INDEX}" -E "${ENV}"
```
roxell, line 72: So for me to add multiple arbitrary environment variables I have to do:
ENV: "argh=var -E argh2=var2 -E arg3=var3" is that correct?

For constancy, can't we remove '-E "${ENV}"' and just have "${ENV}" here and then the user can specify
ENV: "-E argh=var -E argh2=var2 -E arg3=var3" thoughts?

crazoes: I have tried the way you suggested but due to how bash expands variables it doesn't seem to work if you alter the step to look like `- ./kselftest.sh ${ARGS}`, spaces in `ARGS` dont get preserved and if you do `"${ARGS}" it will all get treated as single arg by the script. 

If the user wants to provide additional `-E` arguments they need to do it to the step command with a different parameter name like:

```
params:
  ENV_1: "Value 1"
  ENV_2: "Value 2"
  ENV_3: "Value 3"

run:
  steps:
    - ./kselftest.sh -E "${ENV_1}" -E "${ENV_2}" -E "${ENV_3}"
```

Alternative solution if we want to support multiple environment variables using a single parameter is:
```
params:
  ENV: ENV_1="Value 1" ENV_2="Value 2" ENV_3="Value 3"
run:
  steps:
    - ./kselftest.sh -E "${ENV_1}"
```

and in the `kselftest.sh` we can directly have the following:
```
        E)
            if [ -n "${OPTARG}" ]; then
                echo "executing 'eval ${OPTARG}'"
                eval "export ${OPTARG}"
            fi
            ;;
```

roxell: Reading the help text for ENV I was under the impression that you wanted to support multiple ENVs. If now, maybe we should just reword the helptext a bit to support 1 ENV only ?

roxell: That looks good, thank you @crazoes

roxell (approved):

PR #520, opened 2024-07-08
add module testing

roxell:
Test to load/unload modules multiple times to see if something bad happens, which it shouldn't.

The majority of drivers registers themselves to an API that calls back into the driver when the hardware side (or network protocol, file system, etc) is getting detected or requested.

Loading and unloading modules that the DUT don't have hardware for is likely to find more of the obvious bugs.

### DIFF
--- a/automated/linux/modules/modules.yaml
+++ b/automated/linux/modules/modules.yaml
@@ -1,0 +1,40 @@
+metadata:
+    format: Lava-Test Test Definition 1.0
+    name: module-tests
+    description: |
+                 Load and unload kernel modules.
+    maintainer:
+        - anders.roxell@linaro.org
+    os:
+        - debian
+        - ubuntu
+        - openembedded
+    scope:
+        - functional
+    devices:
+        - rockpi4
+        - x86
+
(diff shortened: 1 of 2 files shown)


automated/linux/modules/modules.yaml
```diff
@@ -0,0 +1,40 @@
...
+    maintainer:
+        - anders.roxell@linaro.org
+    os:
+        - debian
+        - ubuntu
+        - openembedded
+    scope:
+        - functional
+    devices:
+        - juno
```
bhcopeland, line 15: Maybe rockpi instead?

automated/linux/modules/modules.yaml
```diff
@@ -0,0 +1,40 @@
+metadata:
+    format: Lava-Test Test Definition 1.0
+    name: module-tests
+    description: |
+                 Load and unload kernle modules.
```
bhcopeland, line 5: typo

nareshkamboju:
@roxell  Thanks for addressing all the review comments.

PR #523, opened 2024-08-28
automated: linux: fix lmp-device register error message matching

mwasilew:
Backend error message has changed and the test no longer matches the proper string. This patch fixes the issue


mwasilew:
Tested here: https://lava.infra.foundries.io/results/50818/0_lmp-device-register-hsm

roxell (approved):

PR #524, opened 2024-09-09
rt-tests: allow to run threads on all cores

igaw:
`cyclideadline`, `cyclictest` and `signaltest` are able to run a test thread per core. Extend the wrapper script so that it's possible to run a thread an all available cores as second default option. Any other setting e.g. run 2 threads on the last two CPUs is hardware/test setup dependent.

One other maybe sane default would be to run threads an all but the first CPU which acts as a housekeeping CPU. But let's keep it simple and see if this is actually a real use case.


igaw:
Oops, these are  already in. I went through my backlog and didn't realize I already submitted them. Please ignore.

PR #526, opened 2024-09-11
linux: ethernet: Fix grep parameter case for busybox

stefan-muc:
The -A parameter has to be uppercase to work with busybox. As far as I can see it is uppercase in the help of different Linux, but seems to work in lowercase too - but not in busybox.


roxell (commented):
Thank you for the patch.
Can you please reformat your changelog to add linebreak's after 72 characters and please add your signed-off-by

PR #527, opened 2024-09-12
Add sysfs-interface-test

vishalbhoj:


mwasilew:
@vishalbhoj nice job!
I think it might be a good idea to allow users to select which tests to run on their platform. For example the test will always fail on platform without BT and not all boards have BT. Also not all kernels enable BT.

automated/linux/peripherals/sysfs-interface-test.yaml
```diff
@@ -0,0 +1,22 @@
+metadata:
+    format: Lava-Test Test Definition 1.0
+    name: module-tests
```
nareshkamboju, line 3: please change test
name: sysfs-interface-test

automated/linux/peripherals/sysfs-interface-test.yaml
```diff
@@ -0,0 +1,22 @@
...
+                 USB Gadget Framwork test
+    maintainer:
+        - anders.roxell@linaro.org
+    os:
+        - debian
+        - ubuntu
+        - openembedded
+    scope:
+        - functional
+    devices:
```
nareshkamboju, line 14: devices:
Please add known good LAVA connected devices which supports any of these test cases.

automated/linux/peripherals/sysfs-interface-test.sh
```diff
@@ -0,0 +1,204 @@
...
+        echo "Checking connected displays..."
+        xrandr_output=$(xrandr --listmonitors | grep 'Monitors')
+        if [ -n "$xrandr_output" ]; then
+            echo "Connected displays:"
+            xrandr --listmonitors | grep -v 'Monitors'
+        else
+            echo "No displays connected."
+        fi
+    else
+        echo "drm-sysfs-test fail" >> $RESULT_FILE
```
nareshkamboju, line 182: do we have any scope to skip test cases ?
When Kconfigs not enabled / device do not have these peripherals / interfaces we could skip those tests.

nareshkamboju (changes requested):
@vishalbhoj  Thanks taking your time in writing a new test cases.
I have a few minor comments.

vishalbhoj:
@mwasilew Thank you for the feedback. I was just experimenting on the testcase. I'll be closing the PR for now. I'll create a new one with once I have improved the testcase.

automated/linux/peripherals/sysfs-interface-test.sh
vishalbhoj: Thank you @nareshkamboju for the feedback. I'll close the PR for now. I'll open the PR again later once I have improved the implementation.

PR #533, opened 2024-10-16
automated: linux: add video output comparison test

mwasilew:
The test takes a screenshot from video output and compares to the reference image. It is intended to test HDMI output from DUT.

### DIFF
--- a/automated/linux/video/Dockerfile
+++ b/automated/linux/video/Dockerfile
@@ -1,0 +1,4 @@
+FROM python:3
+
+COPY requirements.txt /home/
+RUN python3 -m pip install opencv-python scikit-image requests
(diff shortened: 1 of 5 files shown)


mwasilew:
@roxell it looks like the linter check fail is not related to this patch. Could you take a look?

automated/linux/video/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM python:3
+
+COPY requirements.txt /home/
+RUN python3 -m pip install opencv-python scikit-image requests
```
bhcopeland, line 4: why don't you have all this in requirements.txt and install from there?

mwasilew: It takes 1.5 hours. Both Dockerfile and requirements.txt are not needed for the test. They're added if someone would like to a) build the container to run the test in LAVA, b) run the test locally.

mwasilew: ah, and using requirments.txt in Dockerfile will be tricky. If the requirements.txt changes, the layer cache will not be invalidated.

roxell (approved):

PR #534, opened 2024-10-28
noninteractive-tradefed: enable skipping of internet access check

liuyq:
For some cases that before the WIFI or ethernet features work on the board,
it would be better to be able to run the cts/vts tests for such devboards too,
regardless of the network related failures.
So adding the INTERNET_ACCESS variable to specify
it it's necessary for the test, default to true
for backwards compatibility

### DIFF
--- a/automated/android/noninteractive-tradefed/tradefed.sh
+++ b/automated/android/noninteractive-tradefed/tradefed.sh
@@ -23,6 +23,9 @@ FAILURES_PRINTED="0"
 AP_SSID=""
 # WIFI AP KEY
 AP_KEY=""
+# default to set the wifi(when required) and check the internet when not specified or set to true
+# which means it needs to specify explicitly to false to not check the internet access
+INTERNET_ACCESS="true"
 
 check_internet_access() {
     if [ -n "${AP_SSID}" ] && [ -n "${AP_KEY}" ]; then
@@ -75,11 +78,11 @@ check_internet_access() {
 }
 
 usage() {
-    echo "Usage: $0 [-o timeout] [-n serialno] [-c cts_url] [-t test_params] [-p test_path] [-r <aggregated|atomic>] [-f failures_printed] [-a <ap_ssid>] [-k <ap_key>]" 1>&2
+    echo "Usage: $0 [-o timeout] [-n serialno] [-c cts_url] [-t test_params] [-p test_path] [-r <aggregated|atomic>] [-f failures_printed] [-a <ap_ssid>] [-k <ap_key>] [ -i [true|false]]" 1>&2
     exit 1
 }
 
-while getopts ':o:n:c:t:p:r:f:a:k:' opt; do
+while getopts ':o:n:c:t:p:r:f:a:k:i:' opt; do
     case "${opt}" in
         o) TIMEOUT="${OPTARG}" ;;
         n) export ANDROID_SERIAL="${OPTARG}" ;;
@@ -90,6 +93,7 @@ while getopts ':o:n:c:t:p:r:f:a:k:' opt; do
         f) FAILURES_PRINTED="${OPTARG}" ;;
         a) AP_SSID="${OPTARG}" ;;
         k) AP_KEY="${OPTARG}" ;;
+        i) INTERNET_ACCESS="${OPTARG}" ;; # if check the internet access
         *) usage ;;
     esac
 done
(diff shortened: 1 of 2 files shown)


automated/android/noninteractive-tradefed/tradefed.sh
```diff
@@ -23,13 +23,17 @@ FAILURES_PRINTED="0"
 AP_SSID=""
 # WIFI AP KEY
 AP_KEY=""
+# default to set the wifi(when required) and check the internet when not specified or set to true
+# which means it needs to specify explicitly to false to not check the internet access
```
bhcopeland, line 27: # By default, WiFi setup and internet access checks are enabled (INTERNET_ACCESS="true").
# To disable internet access checks, set INTERNET_ACCESS="false"

automated/android/noninteractive-tradefed/tradefed.sh
```diff
@@ -82,31 +87,33 @@ if [ -e "${TEST_PATH}/testcases/vts/testcases/kernel/linux_kselftest/kselftest_c
...
+        # try to connect to wifi with the external AdbJoinWifi apk from
+        # https://github.com/steinwurf/adb-join-wifi
+        # if AP_SSID and AP_KEY are not specified for this tradefed test action
+        adb_join_wifi "${AP_SSID}" "${AP_KEY}"
+    fi
 
-# wait for a while till the wifi connecting operation finished
-sleep 60
+    # wait for a while till the wifi connecting operation finished
+    sleep 60
```
bhcopeland, line 103: white space

automated/android/noninteractive-tradefed/tradefed.sh
```diff
@@ -82,31 +87,33 @@ if [ -e "${TEST_PATH}/testcases/vts/testcases/kernel/linux_kselftest/kselftest_c
...
+    if adb shell 'ping -c 10 '"${SERVER}"'; echo exitcode: $?' | grep -q "exitcode: 0"; then
+        report_pass "network-available"
+    else
+        report_fail "network-available"
+        # print more debug information
+        adb shell ip address || true
+        adb shell ping -c 10 8.8.8.8 || true
+        # to be caught by the yaml file
+        exit 100
+    fi
```
bhcopeland, line 116: Maybe it would make sense here to have "BACKUP_SERVER" variable? 

BACKUP_SERVER="8.8.8.8"

and then

if adb shell "ping -c 5 ${SERVER}" > /dev/null 2>&1; then
....
else
    if adb shell "ping -c 5 ${BACKUP_SERVER}" > /dev/null 2>&1; then
...
else
<debug info>

automated/android/noninteractive-tradefed/tradefed.sh
liuyq: Thanks!

automated/android/noninteractive-tradefed/tradefed.sh
liuyq: hmm, It's white space here on my local side. Or you mean to change to use tab?

automated/android/noninteractive-tradefed/tradefed.sh
liuyq: ping 8.8.8.8 is to check if there is any DNS issue in the lab. When ping www.google.com failed, but ping 8.8.8.8 passed, then it means there is some DNS issue in the lab, but it's still the network failure.

bhcopeland: Thanks Yongqin. Sorry wasn't clear, I meant to use DNS_TEST_SERVER or BACKUP_SERVER as a variable at the top. I would also add some additional debug info in, when it does fail. Something like this:

```
SERVER="www.google.com"
DNS_TEST_SERVER="8.8.8.8"

info_msg "device-${ANDROID_SERIAL}: Checking network connectivity by pinging ${SERVER}..."

# Try pinging www.google.com first
if adb shell "ping -c 10 ${SERVER}" > /dev/null 2>&1; then
    report_pass "network-available"
else
    info_msg "Ping to ${SERVER} failed, testing DNS by pinging ${DNS_TEST_SERVER}..."

    # If ping to 8.8.8.8 succeeds, it's likely a DNS issue
    if adb shell "ping -c 10 ${DNS_TEST_SERVER}" > /dev/null 2>&1; then
        report_fail "network-available: DNS issue detected"
        
        # DNS-specific debug information
        info_msg "DNS resolution issue suspected; gathering DNS configuration information..."
        adb shell getprop | grep dns || true  # Display DNS settings

    else
        report_fail "network-available: general network failure"
        
        # General network debug information
        info_msg "Network connectivity issue detected; gathering debug information..."
        adb shell ip address || true
        adb shell route || true
        adb shell ifconfig || true
    fi
    
    # Exit with error to trigger YAML file handling
    exit 100
fi

```

liuyq: thanks, that looks really good, let's submit it as another PR instead of being mixed here.

roxell:
@liuyq what would happen if you just dropped "exit 100"?
Then the suite would be running if the network isn't up.

liuyq:
> @liuyq what would happen if you just dropped "exit 100"? Then the suite would be running if the network isn't up. 

For most of the cases, it's expected to run the tests with the internet access. And for such cases if there is any reason the network is not available, the tests would be run without the internet access and report many confusing failures, which is something we want to avoid. Thus the network check is added here. 

For some special cases, we have to run tests without the network support, that's the purpose of this PR, and the existing network check logic should not be broken.

roxell (approved):

roxell:
Thank you @liuyq for your work.

PR #535, opened 2024-10-29
noninteractive-tradefed: improvement on network check

liuyq:
noninteractive-tradefed: improvement on network check

The improvements are using top variables, reporting dns failures
separately, and adding some more meaningful debug output,
as well as the output of ifconfig command. Move all the internet access
into a function named check_internet_access().

Suggested-by: Anders Roxell <anders.roxell@linaro.org>
Suggested-by: Benjamin Copeland <benjamin.copeland@linaro.org>

### DIFF
--- a/automated/android/noninteractive-tradefed/tradefed.sh
+++ b/automated/android/noninteractive-tradefed/tradefed.sh
@@ -24,6 +24,56 @@ AP_SSID=""
 # WIFI AP KEY
 AP_KEY=""
 
+check_internet_access() {
+    if [ -n "${AP_SSID}" ] && [ -n "${AP_KEY}" ]; then
+        # try to connect to wifi with the feature provided by tradefed by default
+        # when AP_SSID and AP_KEY are specified for this tradefed test action explicitly
+        TEST_PARAMS="${TEST_PARAMS} --wifi-ssid ${AP_SSID} --wifi-psk ${AP_KEY}"
+    else
+        # try to connect to wifi with the external AdbJoinWifi apk from
+        # https://github.com/steinwurf/adb-join-wifi
+        # if AP_SSID and AP_KEY are not specified for this tradefed test action
+        adb_join_wifi "${AP_SSID}" "${AP_KEY}"
+    fi
+
+    # wait for a while till the wifi connecting operation finished
+    sleep 60
+
+    SERVER="www.google.com"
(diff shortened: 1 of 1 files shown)


bhcopeland:
LGTM. Small request but "suggested-by" might be a good use here in commit message :) Thanks for the shout-out in the description.

liuyq:
hi, @bhcopeland 
please help check it again, if you could merge this, 
I will rebase https://github.com/Linaro/test-definitions/pull/534 based on this.

automated/android/noninteractive-tradefed/tradefed.sh
```diff
@@ -97,15 +97,36 @@ fi
...
-    adb shell ip address || true
-    adb shell ping -c 10 8.8.8.8 || true
-    # to be caught by the yaml file
+
+    # If ping to $DNS_TEST_SERVER succeeds, it's likely a DNS issue
+    if adb shell 'ping -c 10 '"${DNS_TEST_SERVER}"'; echo exitcode: $?' | grep -q "exitcode: 0"; then
+        report_pass "network-ping-dns-server"
+
+        # DNS-specific debug information
+        info_msg "DNS resolution issue suspected; gathering DNS configuration information..."
```
vishalbhoj, line 116: It is not clear what this message means here. If ping passes, why do we need to print such message ? and Should we print DNS details in case of failure case on DUT as well as on the host ?

liuyq: the ping at L112 is to ping 8.8.8.8, if it passed while the previous ping domain name(www.google.com) failed, then it should indicate some dns setting issue there. The host(and should be worker) dns setting might be different from the DUT side settings, and the host side should work as far as the tests run to here, but saying that, it should be harmless or helpful for debugging on such failed cases. If you have any suggestions on what host side commands to be run, please let me know.

roxell (approved):

bhcopeland (approved):

automated/android/noninteractive-tradefed/tradefed.sh
roxell: nslookup http://google.com/ might work on the host side.

roxell: I'm fine leaving this for a followup PR.

roxell:
Thank you for all the work with this PR @liuyq

PR #540, opened 2024-12-03
automated: linux: xfstests: fix result output

roxell:
The result output doesn't drop '.notrun', '.out.bad' and '.full' from the test name so it looks like 'brtf-006.notrun skip' and the sed cmdline assumes that '.notrun' is at the end of the row, which it were before it got reformated.  When dropping the '$' from sed cmdline it works as expected.


TheodoreRTG (approved):
LGTM

PR #541, opened 2024-12-03
automated: linux: gpiod: parse-output: drop 'tests' prefix

roxell:
The output of the test name always starts with a '_' which is wrong. Removing that by remove the prefixed '{tests}_' the testname, and the 'tests' variable was assigned to an empty string. This resulted in that the testname started with a prefix '_'.


TheodoreRTG (approved):
LGTM

TheodoreRTG (approved):

PR #542, opened 2024-12-09
noninteractive-tradefed: disable ENABLE_XTS_DYNAMIC_DOWNLOADER by default

liuyq:
as suggested by google to avoid unexpected failures like the libcore java.lang.NoClassDefFoundError failures


roxell:
So if I understand correctly this will stop the tests from dynamically download new libraries?
Do we know which version that is the problematic one?

liuyq:
> So if I understand correctly this will stop the tests from dynamically download new libraries?

Yes, you are right, and that's what we want, and it have been confirmed with google.

> Do we know which version that is the problematic one?
Currently only Android15 and AOSP main are reported.

roxell (approved):

PR #546, opened 2025-01-14
automated: linux: add alsa-smoke test

mwasilew:
The test uses alsa-utils tools aplay and arecord to find whether any playback and record devices are available in the system. If the test runs on an OS without package manager and no alsa-utils available it will fall back to look for character devices named "controlCX" where X is a number. This is the same approach that alsa-utils uses. The default path to look for devices is /dev/snd/

### DIFF
--- a/automated/linux/alsa-smoke/alsa-smoke.sh
+++ b/automated/linux/alsa-smoke/alsa-smoke.sh
@@ -1,0 +1,65 @@
+#!/bin/sh
+
+# shellcheck disable=SC1091
+. ../../lib/sh-test-lib
+OUTPUT="$(pwd)/output"
+RESULT_FILE="${OUTPUT}/result.txt"
+export RESULT_FILE
+REFERENCE_PATH="/dev/snd"
+SKIP_INSTALL="False"
+
+usage() {
+    echo "Usage: $0 [-s <true|false>] [-p </path/to/snd/devices>]" 1>&2
+    exit 1
+}
+
+while getopts "s:p:h" o; do
+  case "$o" in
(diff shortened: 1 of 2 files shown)


automated/linux/alsa-smoke/alsa-smoke.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/sh
+
+# shellcheck disable=SC1091
+. ../../lib/sh-test-lib
+OUTPUT="$(pwd)/output"
+RESULT_FILE="${OUTPUT}/result.txt"
+export RESULT_FILE
+REFERENCE_PATH="/dev/snd"
+
```
roxell, line 9: SKIP_INSTALL should be set her to "False"

automated/linux/alsa-smoke/alsa-smoke.sh
```diff
@@ -0,0 +1,64 @@
...
+usage() {
+    echo "Usage: $0 [-s <true|false>] [-p </path/to/snd/devices>]" 1>&2
+    exit 1
+}
+
+while getopts "s:p:d:h" o; do
```
roxell, line 15: 'd' isn't used.

mwasilew: Fixed. I used it for something in one of the previous versions, but now I forgot what it was.

automated/linux/alsa-smoke/alsa-smoke.sh
mwasilew: Fixed.

roxell (approved):

PR #554, opened 2025-03-24
automated: linux: kselftest: publish kselftest failure logs

TheodoreRTG:
Currently kselftest logs for failures aren't displaying full logs.
This change creates log files for each failure and tracks the results.

Forks changes to send-to-lava from PR #552 .

### DIFF
--- a/automated/linux/kselftest/parse-output.py
+++ b/automated/linux/kselftest/parse-output.py
@@ -40,18 +40,24 @@ def parse_nested_tap(string):
 
     output = ""
     ps = parser.Parser()
+    test_name = None
     for l in ps.parse_text(string):
         if l.category == "test":
+            test_name = make_name(l.description, l.directive.text, l.ok, l.skip)
             results.append(
                 {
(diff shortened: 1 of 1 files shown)


automated/linux/kselftest/parse-output.py
```diff
@@ -74,9 +80,19 @@ def make_names_unique(results):
             r["name"] += f"_dup{namecounts[name]}"
 
 
+def make_log_files(results):
+    for r in results:
+        name = r["name"]
+        if r["result"] == "fail":
+            log_file = open(f"output/{name}.log", "w")
```
mwasilew, line 87: it would be a good idea to catch any IO exceptions here. IIRC this script is run on DUT and you may run out of disk space.

TheodoreRTG: Thanks for the comment, I added an exception block. Let me know if there are any issues.

terceiro: should be OK now.

roxell (approved):

PR #555, opened 2025-03-28
noninteractive-tradefed: add java path support for aarch64

liuyq:
The java path on aarch64 is something like
    /usr/lib/jvm/java-11-openjdk-arm64/bin/java
which is different from the path on x86_64 like
    /usr/lib/jvm/java-11-openjdk-amd64/bin/java

change to use differnt paths for different architectures
accordingly to support the aarch64 host as well.

### DIFF
--- a/automated/android/noninteractive-tradefed/setup.sh
+++ b/automated/android/noninteractive-tradefed/setup.sh
@@ -5,10 +5,15 @@
 . ../../lib/sh-test-lib
 . ../../lib/android-test-lib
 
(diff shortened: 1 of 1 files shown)


mwasilew (approved):

bhcopeland:
LGTM. Your commit message, though, is unclear. Can you write a slightly more descriptive reason for the change?

liuyq:
> LGTM. Your commit message, though, is unclear. Can you write a slightly more descriptive reason for the change?

Just updated, please help to check it again

bhcopeland (approved):

PR #557, opened 2025-04-06
automated: linux: Add ALSA basic audio test

broonie:
This is an incredibly basic test, shipped as part of alsa-utils, which
plays and records a tone then uses a FFT to verify that the tone appears
sufficiently cleanly in the output.  It requires that the system under
test have previously been set up with a loopback audio path, either
within the card or via cables.

### DIFF
--- a/automated/linux/alsa-bat/bat.sh
+++ b/automated/linux/alsa-bat/bat.sh
@@ -1,0 +1,72 @@
+#!/bin/sh -e
+# shellcheck disable=SC1091
+
+OUTPUT="$(pwd)/output"
+RESULT_FILE="${OUTPUT}/result.txt"
+
+. ../../lib/sh-test-lib
+
+create_out_dir "${OUTPUT}"
+
+PARAMS=
+
+if [ "${TST_CARD}" != "" ]; then
+	PARAMS="${PARAMS} -D${TST_CARD}"
+fi
+
+if [ "${TST_CHANNELS}" != "" ]; then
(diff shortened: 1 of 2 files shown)


automated/linux/alsa-bat/bat.yaml
```diff
@@ -0,0 +1,47 @@
+metadata:
+    name: meminfo
```
bhcopeland, line 2: This should be alsa-bat

automated/linux/alsa-bat/bat.sh
```diff
@@ -0,0 +1,72 @@
...
+
+if [ "${TST_FORMAT}" != "" ]; then
+	PARAMS="${PARAMS} -f${TST_FORMAT}"
+fi
+
+if [ "${TST_RATE}" != "" ]; then
+	PARAMS="${PARAMS} -r${TST_RATE}"
+fi
+
+if [ "${TST_LENGTH}" != "" ]; then
```
bhcopeland, line 37: Is this meant to be TST_SIGNAL?

bhcopeland:
Thanks for the PR @broonie, just a few minor comments!

bhcopeland (approved):

mwasilew:
I'm a bit late to comment. You can add the description from the commit message to the "description" field in the yaml file. This way it will be rendered in the docs. Example: https://test-definitions.readthedocs.io/en/latest/automated/linux/el2go/el2go.html

PR #560, opened 2025-04-09
noninteractive-tradefed: using HTTP_CACHE to get kisscache url

liuyq:
After the following lava change:
    https://gitlab.com/lava/lava/-/merge_requests/2734
the kisscache url is exported by LAVA via the HTTP_CACHE variable, it's better to check and use the varible in the scripts as well, to avoid being passed from the LAVA job definition.

### DIFF
--- a/automated/android/noninteractive-tradefed/tradefed.sh
+++ b/automated/android/noninteractive-tradefed/tradefed.sh
@@ -123,7 +123,18 @@ file_name=$(basename "${TEST_URL}")
 if echo "${TEST_URL}" | grep "^http://lkft-cache.lkftlab/" ; then
     NO_PROXY=.lkftlab wget -S --progress=dot:giga "${TEST_URL}" -O "${file_name}"
 elif echo "${TEST_URL}" | grep "^http" ; then
-    wget -S --progress=dot:giga "${TEST_URL}" -O "${file_name}"
+    # using kisscache to download the file, based on the following change:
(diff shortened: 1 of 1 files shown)


automated/android/noninteractive-tradefed/tradefed.sh
```diff
@@ -123,7 +123,13 @@ file_name=$(basename "${TEST_URL}")
 if echo "${TEST_URL}" | grep "^http://lkft-cache.lkftlab/" ; then
     NO_PROXY=.lkftlab wget -S --progress=dot:giga "${TEST_URL}" -O "${file_name}"
 elif echo "${TEST_URL}" | grep "^http" ; then
-    wget -S --progress=dot:giga "${TEST_URL}" -O "${file_name}"
+    ### using kisscache to download the file, based on the following change:
+    ###    https://gitlab.com/lava/lava/-/merge_requests/2734
+    if [ -n "${HTTP_CACHE}" ]; then
```
chase-qi, line 128: The setting is exposed directly. It is this format: `https://cache.lavasoftware.org/api/v1/fetch/?url=%s`. I think you need to replace the `%s` with `TEST_URL` here.

liuyq: Thanks for pointing it out! updated to remove %s first.

chase-qi:
LGTM.

bhcopeland (approved):

PR #561, opened 2025-04-11
automated: linux: ltp: kirk runner: parallelize tests

roxell:
Add support for parallel test execution with the new LTP test runner, kirk.  This update introduces the KIRK_WORKERS variable, which allows you to configure the number of worker processes used for running tests in parallel. This can significantly reduce test runtime on systems with many CPU cores.

Suggested-by: Mark Brown <broonie@kernel.org>

### DIFF
--- a/automated/linux/ltp/ltp.sh
+++ b/automated/linux/ltp/ltp.sh
@@ -31,6 +31,7 @@ SHARD_NUMBER=1
 SHARD_INDEX=1
 
 RUNNER=""
+KIRK_WORKERS=1
 
 LTP_TMPDIR=/ltp-tmp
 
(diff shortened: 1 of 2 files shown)


automated/linux/ltp/ltp.sh
```diff
@@ -31,6 +31,7 @@ SHARD_NUMBER=1
 SHARD_INDEX=1
 
 RUNNER=""
+KIRK_WORKERS=1
```
bhcopeland, line 34: Just a thought. Some nproc auto detection here might be useful as well being able to define a number

nareshkamboju (approved):

PR #563, opened 2025-04-21
android: add the Android KUnit test

liuyq:
with the tests.zip package provided, which is generated by the normal gki kernel build

### DIFF
--- a/automated/android/kunit/kunit.sh
+++ b/automated/android/kunit/kunit.sh
@@ -1,0 +1,146 @@
+#!/bin/bash -ex
+
+# shellcheck disable=SC1091
+. ../../lib/sh-test-lib
+DIR_OUTPUT="$(pwd)/output"
+mkdir -p "${DIR_OUTPUT}"
+RESULT_FILE="${DIR_OUTPUT}/result.txt"
+export RESULT_FILE
+
+# shellcheck disable=SC1091
+. ../../lib/android-test-lib
+
+RETRY_COUNT=5
+RETRY_INTERVAL=2
+
+TESTS_ZIP_URL=""
+SQUAD_UPLOAD_URL=""
+TRADEFED_PREBUILTS_GIT_URL="https://android.googlesource.com/platform/tools/tradefederation/prebuilts"
+
+F_TESTS_ZIP="$(pwd)/tests.zip"
+DIR_TESTS="$(pwd)/tests"
+DIR_TEST_LOGS="${DIR_OUTPUT}/test-logs"
+F_KUNIT_LOG="${DIR_TEST_LOGS}/kunit.log"
+DIR_TF_PREBUILTS="$(pwd)/prebuilts"
+
+function usage(){
+    echo "Usage: $0 -u <TESTS_ZIP_URL> [ -s <SQUAD_UPLOAD_URL>]" 1>&2
+    exit 1
+}
+
+function upload_logs_to_squad(){
+    if [ -z "${SQUAD_UPLOAD_URL}" ]; then
+        return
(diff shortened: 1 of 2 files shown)


automated/android/kunit/kunit.sh
```diff
@@ -0,0 +1,134 @@
...
+prebuilts/filegroups/tradefed/tradefed.sh  \
+    run commandAndExit \
+    template/local_min \
+    --template:map test=suite/test_mapping_suite \
+    --include-filter kunit \
+    --tests-dir="${DIR_TESTS}"  \
+    --log-file-path="${DIR_TEST_LOGS}" \
+    -s "${ANDROID_SERIAL}" |& tee "${F_KUNIT_LOG}"
+
+parse_kunit_log "${F_KUNIT_LOG}"
```
mwasilew, line 132: you might want to do that in the postprocessing rather than in LAVA job. It dependes on how many results there are. If the number isn't too high, parsing in LAVA job is fine.

mwasilew (commented):
LGTM. I added a comment about possible performance issue when parsing results in a LAVA job. Otherwise I see no issues.

automated/android/kunit/kunit.sh
liuyq: Thanks for the check! Yes, that's one of my concern as well, but consider currently there are only about 500 test case, it should be OK, maybe it's still OK after 10 times increased.

mwasilew: 500 should be OK. I wouldn't go beyond 1k. LAVA is very slow when it comes to recording the results from the test job.

automated/android/kunit/kunit.sh
```diff
@@ -0,0 +1,134 @@
...
+# download and unzip tests.zip
+rm -f "${F_TESTS_ZIP}" && \
+    curl -fsSL "${TESTS_ZIP_URL}" -o "${F_TESTS_ZIP}"
+rm -fr "${DIR_TESTS}" && \
+    mkdir -p "${DIR_TESTS}" && \
+    unzip -o "${F_TESTS_ZIP}" -d "${DIR_TESTS}"
+
+# clone the tradefed prebuilts repository
+rm -fr "${DIR_TF_PREBUILTS}" && \
+    git clone --depth 1 "${TRADEFED_PREBUILTS_GIT_URL}" "${DIR_TF_PREBUILTS}"
```
bhcopeland, line 119: same here r.e. retry

automated/android/kunit/kunit.sh
```diff
@@ -0,0 +1,134 @@
...
+initialize_adb
+
+if [ -z "${TESTS_ZIP_URL}" ]; then
+    echo "The TESTS_ZIP_URL must be specified."
+    exit 1
+fi
+
+# download and unzip tests.zip
+rm -f "${F_TESTS_ZIP}" && \
+    curl -fsSL "${TESTS_ZIP_URL}" -o "${F_TESTS_ZIP}"
```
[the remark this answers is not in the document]
liuyq: updated with the retry options, please help check again

automated/android/kunit/kunit.sh
liuyq: updated to be done in a while block, please help to check it again.

liuyq:
any idea about the Build docs error? or anyone could help to re-trigger it again ?

mwasilew:
@liuyq let me check. Doesn't seem related to the PR

roxell:
> I can reproduce it with master branch as well. Will post a fix shortly.

Thank you @mwasilew

PR #566, opened 2025-05-07
Update modules

roxell:


automated/linux/modules/modules.sh
```diff
@@ -76,6 +76,11 @@ report() {
 	fi
 }
 
+scan_dmesg_for_errors() {
+	echo "=== Scanning dmesg for errors ==="
+	dmesg -l 0,1,2,3,4,5 | grep -E "BUG:|WARNING:|Oops:|Call Trace:" && report_fail "dmesg_error_scan" || report_pass "dmesg_error_scan"
```
bhcopeland, line 92: dmesg 5 is super verbose, do you think we can get away with a lower number?

bhcopeland, line 92: Could make this case instensive with -i

automated/linux/modules/modules.sh
```diff
@@ -81,6 +81,16 @@ scan_dmesg_for_errors() {
 	dmesg -l 0,1,2,3,4,5 | grep -E "BUG:|WARNING:|Oops:|Call Trace:" && report_fail "dmesg_error_scan" || report_pass "dmesg_error_scan"
 }
 
+check_module_unloaded() {
+	local _module="$1"
+	if lsmod | grep "^${_module}"; then
```
bhcopeland, line 97: Might want to add a awk/cut here to lessen the risk of false positives

automated/linux/modules/modules.sh
roxell: don't think its too bad with 5, we excludes 6 (info) and 7 (debug)

automated/linux/modules/modules.sh
roxell: I'll update to this:
lsmod | grep "^${_module} " > /dev/null

bhcopeland (approved):

PR #570, opened 2025-06-03
automated: linux: add tcpreplay testing framework

roxell:
This patch introduces an automated testing setup for tcpreplay-based traffic replay and analysis. The system consists of:

1. `tcpreplay.py`: A test runner that:
   - Creates and configures a TAP interface
   - Replays PCAP files using `tcpreplay`
   - Tracks results and produces a summary for each test
   - Handles both expected and unexpected results (e.g., xfail -> pass)

2. `generate_pcap.py`: A Scapy-based script to:
   - Generate a suite of PCAP files for functional and edge case testing
   - Include both valid packets and malformed/false-positive scenarios
   - Provide coverage for multiple protocols (TCP, UDP, ICMP, DNS, etc.)
   - Simulate problematic flows like fragmented packets and invalid flags

Highlights:
- xfail support: tests expected to fail are counted as 'pass' if they do fail
- xpass detection: alerts if a known-broken test unexpectedly succeeds
- Easy extensibility for new PCAPs and expectations

Example test cases include:
- TCP lifecycle (`tcp_basic`, `tcp_full_cycle`)
- Bad flag scenarios (`bad_tcp_flags`)
- Noise/overlap (`false_positive_overlap`)
- Fragmentation and malformed headers

### DIFF
--- a/automated/linux/tcpreplay/pcap/generate_pcap.py
+++ b/automated/linux/tcpreplay/pcap/generate_pcap.py
@@ -1,0 +1,199 @@
+#!/usr/bin/env python3
+# vim: set ts=8 sw=4 sts=4 et tw=80 fileencoding=utf-8 :
+from scapy.all import (
+    Ether,
+    IP,
+    TCP,
+    UDP,
+    ICMP,
+    DNS,
+    DNSQR,
+    DNSRR,
+    wrpcap,
+    Raw,
+    fragment,
+)
+
+
+test_expectations = {
+    "tcp_basic.pcap": "pass",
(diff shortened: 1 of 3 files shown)


nareshkamboju (approved):

roxell:
> @roxell I might be wrong here, but is it really necessary to include all .pcap files? IIUC you include the script to generate them. I would really like to reduce the size of this repository. Adding binaries doesn't help in this quest.

should be fine to generate them, its just one extra python package to include into the rootfs... @mwasilew 
I wasn't sure but I'm fine changing that.

PR #580, opened 2025-07-11
noninteractive-tradefed: support Android16 and the latest branches

liuyq:
which requires to use JDK 21  to run the cts/vts packages. Otherwise the following error will be reported with JDK 17:
    java.lang.UnsupportedClassVersionError: \
        com/android/tradefed/targetprep/MultiSimPreparer has been compiled \
        by a more recent version of the Java Runtime (class file version 65.0), \
        this version of the Java Runtime only recognizes class file versions up to 61.0

Tested on Android11, Android12, Android13, Android14, Android15 and AOSP latest versions with the cts-lkft, vts-kernel tests


automated/android/noninteractive-tradefed/setup.sh
```diff
@@ -11,9 +11,12 @@ if [ "X$(uname -m)" = "Xaarch64" ]; then
     java_path_arch_str="arm64"
 fi
 java_path="/usr/lib/jvm/java-11-openjdk-${java_path_arch_str}/bin/java"
-if [ -n "${ANDROID_VERSION}" ] && echo "${ANDROID_VERSION}" | grep -E -q "aosp-android14|aosp-main"; then
+if [ -n "${ANDROID_VERSION}" ] && echo "${ANDROID_VERSION}" | grep -E -q "aosp-android14|aosp-android15|aosp-main"; then
```
bhcopeland, line 14:
```
get_required_java_version() {
    case "${ANDROID_VERSION}" in
        *android11*|*android12*|*android13*) echo "11" ;;
        *android14*|*android15*|*aosp-main*) echo "17" ;;
        *android16*|*aosp-latest*) echo "21" ;;
        *) echo "21" ;;  
    esac
}

java_version=$(get_required_java_version)
java_path="/usr/lib/jvm/java-${java_version}-openjdk-${java_path_arch_str}/bin/java"
```

Just thinking about this, wouldn't this approach be slightly easier to manage? We can have a fallback to 21, and by having a lookup table approach will be easier to manage.

liuyq: Thanks for the suggestion! updated already, please help to check again.

bhcopeland (approved):

PR #584, opened 2025-08-05
linux: perf: Run test sequentially

Leo-Yan:
CI reports random failures for perf test and may be unstable when tests are executed in parallel. Enforce sequential test execution.

### DIFF
--- a/automated/linux/perf/perf.sh
+++ b/automated/linux/perf/perf.sh
@@ -11,19 +11,23 @@ TEST_FAIL_LOG="${OUTPUT}/test_fail_log.txt"
 TEST_SKIP_LOG="${OUTPUT}/test_skip_log.txt"
 export RESULT_FILE
 SKIP_INSTALL="false"
+PARALLEL_TESTS="false"
 # List of test cases
 TEST="record report stat test"
 # PERF version
 PERF_VERSION="$(uname -r | cut -d . -f 1-2)"
 
 usage() {
-    echo "Usage: $0 [-s <true|false>]" 1>&2
+    echo "Usage: $0 [options...]" 1>&2
+    echo " [-s <true|false>]        Skip to install perf" 1>&2
+    echo " [-p <true|false>]        Run tests in parallel" 1>&2
     exit 1
 }
 
-while getopts "s:h" arg; do
+while getopts "s:p:h" arg; do
    case "$arg" in
      s) SKIP_INSTALL="${OPTARG}";;
+     p) PARALLEL_TESTS="${OPTARG}";;
      h|*) usage ;;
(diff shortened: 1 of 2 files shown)


Leo-Yan:
Hi @nareshkamboju , any concern for this patch or is it fine for merging it?

automated/linux/perf/perf.sh
```diff
@@ -81,7 +83,11 @@ run_perf_stat() {
 run_perf_test() {
     # Test 'perf test'
     info_msg "Performing 'perf test'..."
-    perf test -v 2>&1 | tee "${RESULT_LOG}"
+    if [ "${RUN_TESTS_IN_PARALLEL}" = "False" ] || [ "${RUN_TESTS_IN_PARALLEL}" = "false" ]; then
```
bhcopeland, line 86: Can we just go with false, not "False". Maybe this check should just check for "true". 
```
if [ "${RUN_TESTS_IN_PARALLEL}" = "true" ]; then
<blah>
else
<blah>
fi
```

automated/linux/perf/perf.yaml
```diff
@@ -22,9 +22,10 @@ metadata:
         - lava-test-shell
 params:
     SKIP_INSTALL: false
+    RUN_TESTS_IN_PARALLEL: false
```
bhcopeland, line 25: Maybe call it PARALLEL_TESTS? Seems a bit long :)

automated/linux/perf/perf.sh
```diff
@@ -11,19 +11,21 @@ TEST_FAIL_LOG="${OUTPUT}/test_fail_log.txt"
...
 SKIP_INSTALL="false"
+RUN_TESTS_IN_PARALLEL="false"
 # List of test cases
 TEST="record report stat test"
 # PERF version
 PERF_VERSION="$(uname -r | cut -d . -f 1-2)"
 
 usage() {
-    echo "Usage: $0 [-s <true|false>]" 1>&2
+    echo "Usage: $0 [-s <true|false>] [-p <true|false>]" 1>&2
```
bhcopeland, line 21: Can we new line this? so a echo for each

automated/linux/perf/perf.sh
Leo-Yan: Have updated for this.

automated/linux/perf/perf.sh
Leo-Yan: Yes, the new version has changed to:
```
Usage: ./perf.sh [options...]
 [-s <true|false>]        Skip to install perf
 [-p <true|false>]        Run tests in parallel
```

PR #587, opened 2025-08-27
automated: linux: kunit: add tap parser and collect error logs

TheodoreRTG:
This changes parse-output.py in the kunit tests to use tap parser as the logs are in ktap format. Log snippets for tests are located in log files associated with each test name.


TheodoreRTG:
Tested with the following log file:
https://qa-reports.linaro.org/api/testruns/29676487/log_file/

roxell:
Can we add something with 'install_deps "python3-tap" "${SKIP_INSTALL}"' if the dependency isn't installed we can install it.

roxell (approved):

bhcopeland (approved):

roxell:
I forgot to mention, I ran a qemu-arm64 run with and without the change and it wasn't any difference in time it took to parser the results.

PR #591, opened 2025-09-15
fix: add ssh variables to def params

tynorantoni:
Add missing ssh env vars to params so they could be found on lava integration test jobs


stevanradakovic:
lgtm

PR #624, opened 2026-04-20
automated: linux: fix sample file download in network-basic

mwasilew:
URL for downloading sample file was hardcoded in the network-basic test. Since the URL became unavailable, the test started failing. This patch fixes the issue of default URL (set it do debian installer file) and provide means to update the URL in the future, should the problem with hosted file happen again.


mwasilew:
@bhcopeland this is the fix for the issue we discussed today morning (offline samplemedia.linaro.org). Run a test here: https://lava.infra.foundries.io/scheduler/job/183654#L1840

roxell (approved):

PR #625, opened 2026-04-24
automated: linux: kselftest: Don't skip the main rseq tests

broonie:
The rseq run_params_test.sh runner is currently skipped, according to
the changelog due to it being very slow (the comments in the file don't
agree but appear to be bitrotted).  While it is slow it's the main test
here and does work, remove the skip since this is more a policy decision
for individual testers than something that's actively broken.

### DIFF
--- a/automated/linux/kselftest/skipfile-lkft.yaml
+++ b/automated/linux/kselftest/skipfile-lkft.yaml
@@ -82,24 +82,6 @@ skiplist:
       - mqueue:mq_perf_tests
 
   - reason: >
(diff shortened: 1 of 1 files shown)


bhcopeland:
Good spot @broonie. We need to probably re-think that skip list, as its very easy for it to end up in bit-rot state, and its easy for skips tests to make it back into service.

bhcopeland (approved):

broonie:
Yes, I was looking at the rest of the file and thinking it probably needed an audit (there's comments about v4.0 in there!).

PR #630, opened 2026-07-01
Fixes for MMTests scripts

manul7:


automated/linux/mmtests/collector.py
roxell: shouldn't this try except block happen before the the two sys.exit(1) ?

roxell (approved):

PR #631, opened 2026-07-23
automated: linux: wlan-smoke: test all interfaces

lool:
Linux/systemd's predictable network interface naming derives the name from how the WiFi chip is attached: PCIe-attached radios get wlpXsY (bus/slot based, e.g. wlp1s0), while SDIO/platform-integrated radios typically keep the kernel default wlanN. So different devices legitimately end up with different interface names.

Change logic, make DEVICE empty by default which means discovering every interface under /sys/class/net/*/wireless. Then run the smoke tests against each one. This avoids hardcoding a device name in job definitions. Boards without any wireless interface report wlan-device-exists as skip instead of failing the job. Explicit -d/DEVICE still overrides auto-detection for targeting a single interface.


lool:
Tested here: https://lava.infra.foundries.io/scheduler/job/330523

obbardc (approved):
LGTM

automated/linux/wlan-smoke/wlan-smoke-test.sh
```diff
@@ -44,50 +44,68 @@ while getopts "d:b:w:" o; do
...
+        basename "$(dirname "${dev}")"
+    done
+}
+
 # sanity test ip command
 test_iplink() {
-    info_msg "Running ip link test..."
-    ip link show "${DEVICE}"
-    exit_on_fail "ip-link"
+    # shellcheck disable=SC2039
```
fboudra: the 4 inline `# shellcheck disable=SC2039` can be moved to line 24, the one file-level directive: same coverage, cleaner, one source of truth

automated/linux/wlan-smoke/wlan-smoke-test.sh
```diff
@@ -100,11 +118,28 @@ create_out_dir "${OUTPUT}"
...
-test_wlan_down
-test_wlan_up
+
+# use the explicitly requested device, or auto-detect all wireless
+# interfaces present on the DUT
+if [ -n "${DEVICE}" ]; then
+    devices="${DEVICE}"
+else
+    devices="$(list_wlan_devices)"
+fi
```
fboudra: nipick. Same logic, but shorter:
```suggestion
devices="${DEVICE:-$(list_wlan_devices)}"
```

automated/linux/wlan-smoke/wlan-smoke-test.sh
```diff
@@ -44,50 +44,68 @@ while getopts "d:b:w:" o; do
...
 # sanity test ip command
 test_iplink() {
-    info_msg "Running ip link test..."
-    ip link show "${DEVICE}"
-    exit_on_fail "ip-link"
+    # shellcheck disable=SC2039
+    local device="$1"
+    info_msg "Running ip link test on ${device}..."
+    ip link show "${device}"
+    exit_on_fail "ip-link-${device}"
```
bhcopeland: This should use check_return (like the others).

automated/linux/wlan-smoke/wlan-smoke-test.sh
```diff
@@ -44,50 +44,68 @@ while getopts "d:b:w:" o; do
...
     fi
-    check_return "wlan-boot"
+    check_return "wlan-boot-${device}"
 }
 
 # test WLAN device is up
 test_wlan_up() {
-    info_msg "Running wlan-up test..."
-    ip link set "${DEVICE}" up
+    # shellcheck disable=SC2039
```
bhcopeland: Same here R.e. Fathi comment.

automated/linux/wlan-smoke/wlan-smoke-test.sh
```diff
@@ -44,50 +44,68 @@ while getopts "d:b:w:" o; do
...
+    info_msg "Running wlan boot test on ${device}..."
     if [ "${BOOT}" = "enabled" ]; then
-        ip link show "${DEVICE}" | grep "${DEVICE}" | grep "UP"
+        ip link show "${device}" | grep "${device}" | grep "UP"
     else
-        ip link show "${DEVICE}" | grep "${DEVICE}" | grep -v "UP"
+        ip link show "${device}" | grep "${device}" | grep -v "UP"
     fi
-    check_return "wlan-boot"
+    check_return "wlan-boot-${device}"
```
roxell: Can we keep the plain names and put the interface in the info_msg only? If more than one iterface is tested, wrap each interface in the test set, please see report_set_(start|stop) in automated/lib/sh-test-lib.

if not we can't compare the tests anymore.

automated/linux/wlan-smoke/wlan-smoke-test.sh
roxell: right, that is because if two interfaces and the first fails, the second one is never tested and never reported, and the job still looks green.

automated/linux/wlan-smoke/wlan-smoke-test.sh
lool: Done, thanks; note that I see now that SC3043 is disabled globally, so perhaps I should just drop that (bunch of SC2039 in other files though).

automated/linux/wlan-smoke/wlan-smoke-test.sh
lool: Done, thanks!

automated/linux/wlan-smoke/wlan-smoke-test.sh
lool: Ah that's indeed a problem, updated

automated/linux/wlan-smoke/wlan-smoke-test.sh
lool: Ack; done too

automated/linux/wlan-smoke/wlan-smoke-test.sh
lool: Makes sense; updated

fboudra (approved):
LGTM

bhcopeland (approved):

roxell (approved):
Thank you for fixing all the review comments @lool

lool:
> Thank you for fixing all the review comments @lool

Thanks for the thorough reviews!

PR #632, opened 2026-07-30
ltp: add a -r option to select the kirk binary path

bhcopeland:
ltp.sh hardcoded the kirk runner with no way to point at a kirk binary installed outside PATH (for example /opt/kirk/kirk). Rename the variable to KIRK_PATH, add a -r option to set it, and pass it through ltp.yaml. It still defaults to kirk, so existing behaviour is unchanged.

### DIFF
--- a/automated/linux/ltp/ltp.sh
+++ b/automated/linux/ltp/ltp.sh
@@ -30,7 +30,7 @@ BUILD_FROM_TAR="false"
 SHARD_NUMBER=1
 SHARD_INDEX=1
 
-RUNNER="kirk"
+KIRK_PATH="kirk"
 KIRK_WORKERS=1
 EXTRA_KIRK_ARGS=""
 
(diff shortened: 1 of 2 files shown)


automated/linux/ltp/ltp.sh
```diff
@@ -184,20 +190,20 @@ run_ltp() {
     cat runtest/shardfile
     echo "===========End Tests to run ==============="
 
-    if [ -z "${RUNNER}" ] && [ -x "${LTP_INSTALL_PATH}/kirk" ]; then
-        RUNNER="${LTP_INSTALL_PATH}/kirk"
+    if [ -z "${KIRK_PATH}" ] && [ -x "${LTP_INSTALL_PATH}/kirk" ]; then
+        KIRK_PATH="${LTP_INSTALL_PATH}/kirk"
     fi
```
roxell: we can now drop this too, since "-z ${KIRK_PATH}" wont be zero.

bhcopeland: Fixed

roxell (approved):

PR #635, opened 2026-08-25
fix: ltp-realtime: fix python3 shebang and add missing hrtimer_prio parser

Gourav7389:
## Problem
Two issues in `ltp-realtime.py` that cause result reporting to break
on Debian-based targets (bookworm/trixie):

1. **Wrong shebang** (`#!/usr/bin/python`): Debian bookworm/trixie does
   not ship a plain `python` binary, so `ltp-realtime.py` never runs,
   `result.txt` ends up empty, and LAVA only shows a single aggregate
   result instead of per-test results.

2. **Missing `hrtimer_prio` parser**: `hrtimer-prio.c` prints a
   `Result: PASS|FAIL` line but the dispatch table had no entry for it,
   causing a `KeyError` crash that dropped all results for tests running
   after `hrtimer-prio`.

## Fix
- Switch shebang to `#!/usr/bin/env python3`
- Add a minimal `hrtimer_prio()` parser using `result_results()`


mwasilew (approved):

fboudra (approved):
LGTM

bhcopeland:
Can you fix the lint error please?

Gourav7389:
> Can you fix the lint error please?

Fixed! black wanted an extra blank line after `hrtimer_prio()` before the comment block. Thanks.

PR #636, opened 2026-08-26
ci: add dependabot config

bhcopeland:
Track updates for the GitHub Actions used by the workflows, the base images in the Dockerfiles under test/ and automated/linux/, and the pip requirements files.

Action updates are grouped into a single pull request; pip runs monthly since the requirements are unpinned and rarely need attention.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,28 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly
+    commit-message:
+      prefix: ci
+    groups:
+      actions:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,22 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly
```
roxell, line 6: Can we add a prefix to the comimit message like "ci: " ?
Same for the others maybe?

bhcopeland: Updated :)

roxell (approved):

